PR #16, opened 2015-11-03
debian: decrease required cmake version

nschloe:
This corresponds with TriBits's requirement.


nschloe:
I'm preparing a more concise PR.

PR #21, opened 2015-11-05
Update README.md

nschloe:
This fixes various indentations, removes specific dependency versions (since they can change any day; the link to the download page should suffice), and adds a build success badge from Travis.

(Also, This PR should now be automatically tested.)


gdsjaar:
Thanks, looks very good.

PR #33, opened 2016-03-11
Add ability to build the blot_met target

milljm:
re?-add ability to build the blot_met target (not sure when this stopped happening).
Also added a small fix to the blot.in script file. Blot was not detecting the switch from blot.x11 to blot_x11 correctly due to a separator change.

### DIFF
--- a/packages/seacas/applications/blot/CMakeLists.txt
+++ b/packages/seacas/applications/blot/CMakeLists.txt
@@ -42,4 +42,11 @@ TRIBITS_ADD_EXECUTABLE(blot_cps NOEXEPREFIX NOEXESUFFIX
 			        COMM serial mpi)
 install_executable(blot_cps)
 
+TRIBITS_ADD_EXECUTABLE(blot_met NOEXEPREFIX NOEXESUFFIX
+				LINKER_LANGUAGE Fortran
+			        SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/blot.f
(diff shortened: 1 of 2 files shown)


gsjaardema:
Thanks for finding the blot.in separator issue.

Your change to add blot_met really just adds an alias to the blot_cps -- it will output cps files and not met (metafile) files.  I removed the metafile output a few years ago since the majority of people using it would then just convert to postscript and there is a direct postscript output which eliminates the need for the metafile conversion program (post).  The other use of the metafile output was in the dual device blot which would display to X-window and then you could press 'H' or 'hardcopy' to dump that view to a metafile.  There is now the xcps version of blot which does the same thing directly to postscript.

gsjaardema:
Blot script pushed in 8aae4d58f61ab3330b2ae376ff544a7181c52e88

milljm:
Oh, I didn't know about blot_cps being the same. I will pass that info along to the folks that asked me to figure it out :)

gdsjaar:
blot_cps isn't the same, but in the CMakeLists.txt modifications that you made, the blot_met that you were creating would end up being the same as blot_cps and wouldn't be a blot_met that output a metafile...

..Greg

--
"A supercomputer is a device for turning compute-bound problems into I/O-bound problems"

gsjaardema:
Separator issue was merged.  Closing without applying the _met portion of the pull request.

PR #35, opened 2016-03-16
adds some methods to the exodus wrappers for polyhedral elements

ecoon:
Note that I haven't seen any python tests -- if they are there and I missed them, it is likely that they will now break, as the signature of the class __init__ method is now different to allow the extended form of the init_params() struct.

### DIFF
--- a/packages/seacas/scripts/exodus.py.in
+++ b/packages/seacas/scripts/exodus.py.in
@@ -200,22 +200,63 @@ def ex_entity_type(varType):
   # none found, must be invalid`
   return -1 ##EX_INVALID
 
+# init params struct
+class ex_init_params(Structure):
+    """
+    Parameters defining the model dimension, note that many are optional.
+    
+    <int>     num_dim        number of model dimensions
+    <int>     num_nodes      number of model nodes
+    <int>     num_edge       number of model edges
+    <int>     num_edge_blk   number of model edge blocks
+    <int>     num_face       number of model faces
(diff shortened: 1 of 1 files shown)


gsjaardema:
There should be a testExo.py in the SEACAS_Test directory.  More of an integration test than a regression type test.  Not sure we can change the signature of the init method; perhaps have two methods with different signature (not sure if possible in Python) and have the current method fill out the `ex_init_parms` structure and then forward that to the new method. 

We have several users who have existing scripts and can't break them.  Must have backward compatibility.

gsjaardema:
This was merged manually with some additional changes/fixes.

gsjaardema:
Do you have any test or example files using this capability that I could add to the test or example suite?

gsjaardema:
This has been merged manually.

ecoon:
Sorry Greg, I missed your email last week.  Thanks for pulling this in, and
apologies for leaving it half done.

Ethan

On Tuesday, May 17, 2016, Greg Sjaardema <notifications@github.com> wrote:

> This has been merged manually.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/gsjaardema/seacas/pull/35#issuecomment-219840467>
>


-- 

-------------------------------------------------------------------
Ethan Coon

http://www.ldeo.columbia.edu/~ecoon/
-------------------------------------------------------------------

PR #81, opened 2017-09-06
Remove unparsed ignored arguments (TriBITSPub/TriBITS#200)

bartlettroscoe:
This set of changes is needed to remove unparsed ignored arguments in SEACAS.  I already merged these changes to Trilinos/packages/seacas in order to allow the update of TriBITS that now flags these cases.  For details, see:

* https://tribits.org/doc/TribitsDevelopersGuide.html#project-name-check-for-unparsed-arguments

This needs to be merged to the main SEACAS master branch, then snapshotted to the Sierra.base repo copy of SEACAS.  Then when Brent P. snapshots SEACAS from Sierra.base to Trilinos, these changes will be preserved.  Otherwise, these changes will be overwritten and the configure of SEACAS with updated TriBITS will generate several configure-time warnings.


bartlettroscoe:
Thanks!

PR #82, opened 2017-09-26
Fix an error in column deduction in prismatic meshes

raovgarimella:
Two minor bugs were preventing the code from deducing columns in prismatic meshes - instead the column was growing indefinitely with repeated elements. Fixed (tested on some large meshes with about a million elements)


gsjaardema:
Thanks.  Looks good.

PR #84, opened 2017-11-14
Debian fixes

nschloe:
Mostly spelling errors.


gsjaardema:
The zoltan, TriBITs, and kokkos source gets snapshotted in from their respective repositories, so spelling fixes will be overwritten at next snapshot, but the SEACAS fixes should persist.
Thanks for the PR.

PR #87, opened 2017-11-17
Update ex_copy.c

bourdin:
copying a 2D normal model fails right now as coord_z is not defined in the output file.
This patch fixes this issue


gsjaardema:
Thanks.  I guess we don't use 2D much...

PR #93, opened 2018-01-29
ex_utils: compare characters against NUL

mathstuf:
The code had been comparing against a pointer to characters rather than
a character.


mathstuf:
I don't think the CI errors are related. Two have what look like bad CTest commands (or a missing dependency on the system), another had trouble fetching packages from Ubuntu and the last…not really sure.

PR #94, opened 2018-02-08
exodus: support older Visual Studio compilers

mathstuf:
Visual Studio 2013 only supports `__func__` in C++ code and `snprintf`
is available as `_snprintf`.


gsjaardema:
Thanks.

Issue #97, opened 2018-03-09
Remove '#line 1' from flex generated files to avoid bug in nvcc with CMake + Ninja + CUDA builds of Trilinos


bartlettroscoe:
There is a strange defect with `nvcc` when calling through Ninja as generated by CMake described in trilinos/Trilinos#2359 that causes `nvcc` to spit out incorrect dependencies for all line in source files of the form `#line 1 <some-file>`.  This results in unnecessary rebuilds of a bunch of files when ever `ninja` is called to build.  A defect report has been filed with NVIDA but the workaround is to just remove the lines `#line 1 <some-file>` from all of the `flex` generated SEACAS source files.  In particular, with the patch:

```
diff --git a/packages/seacas/libraries/aprepro_lib/apr_scanner.cc b/packages/seacas/libraries/aprepro_lib/apr_scanner.cc
index 747470a..a2512cd 100644
--- a/packages/seacas/libraries/aprepro_lib/apr_scanner.cc
+++ b/packages/seacas/libraries/aprepro_lib/apr_scanner.cc
@@ -971,7 +971,6 @@ static yyconst flex_int16_t yy_rule_linenum[102] = {
 #define yymore() yymore_used_but_not_detected
 #define YY_MORE_ADJ 0
 #define YY_RESTORE_YY_MORE_OFFSET
-#line 1 "/scratch/gdsjaar/seacas/packages/seacas/libraries/aprepro_lib/aprepro.ll"
 /* -*- Mode: c++ -*- */
 #line 39 "/scratch/gdsjaar/seacas/packages/seacas/libraries/aprepro_lib/aprepro.ll"
```

or the patch

```
diff --git a/packages/seacas/libraries/aprepro_lib/apr_scanner.cc b/packages/seacas/libraries/aprepro_lib/apr_scanner.cc
index 747470a..5774992 100644
--- a/packages/seacas/libraries/aprepro_lib/apr_scanner.cc
+++ b/packages/seacas/libraries/aprepro_lib/apr_scanner.cc
@@ -971,7 +971,7 @@ static yyconst flex_int16_t yy_rule_linenum[102] = {
 #define yymore() yymore_used_but_not_detected
 #define YY_MORE_ADJ 0
 #define YY_RESTORE_YY_MORE_OFFSET
-#line 1 "/scratch/gdsjaar/seacas/packages/seacas/libraries/aprepro_lib/aprepro.ll"
[...]

PR #98, opened 2018-03-10
Workaround for ninja+cuda rebuild problems in Trilinos (#97)

bartlettroscoe:
This addressed #97 with a short-term fix (until the next time these files are regenerated).

This remove the line starting wtih '#line 1' which causes nvcc to create a
wrong dependency on apr_scanner.o based on the non-existant file
/scratch/gdsjaar/seacas/packages/seacas/libraries/aprepro_lib/aprepro.ll.  See
trilinos/Trilinos#2359.

This is the same patch as the Trilinos patch in PR trilinos/Trilinos#2360.  By
making this patch here as well, the next time SEACAS is snapshotted into
Trilinos, it will not overwrite the patch in trilinos/Trilinos#2360 being
committed to Trilinos.

But the right solution will be to avoid lines starting with '#line 1' in the
future when generating these files.

### DIFF
--- a/packages/seacas/libraries/aprepro_lib/apr_scanner.cc
+++ b/packages/seacas/libraries/aprepro_lib/apr_scanner.cc
@@ -971,7 +971,6 @@ static yyconst flex_int16_t yy_rule_linenum[102] = {
 #define yymore() yymore_used_but_not_detected
 #define YY_MORE_ADJ 0
 #define YY_RESTORE_YY_MORE_OFFSET
-#line 1 "/scratch/gdsjaar/seacas/packages/seacas/libraries/aprepro_lib/aprepro.ll"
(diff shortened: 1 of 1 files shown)


gsjaardema:
Instead of this PR, I have instead applied the commit 0020417. This eliminates all #line directives in the flex generated files and should ensure they won't be regenerated the next time the aprepro.ll file is changed.  The commit adds the "--noline" directive to the flex options.

bartlettroscoe:
@gsjaardema, Thanks for fixing this!

Are you then okay with removing this line in the Trilinos version of SEACAS to address this issue right now?  If so, can you please commit in https://github.com/trilinos/Trilinos/pull/2360?

Thanks

PR #103, opened 2018-05-02
Quality of life changes and first stab at getting SEACAS cmake to work with Intel and Cray environments and compilers

npe9:


gsjaardema:
I'll take the MPI and intel change, but don't like the deprecated warning change since it eliminates the option to enable/disable deprecated code which is needed.

Can you modify the PR to not delete the `HIDE_DEPRECATED_CODE` symbol.

I guess the .gitignore changes are fine, but they will need to be updated as new versions of the TPLs are released.

npe9:
@gsjaardema Please take another look. Everything should be fixed (including the gitignore version).

PR #104, opened 2018-05-03
Get TPLs working on Cray

npe9:


npe9:
I've run into an interesting problem compiling on the cray nodes. Login nodes can't run H5detect when making the TPLs (illegal instruction because it's compiling for Haswell/KNL and the login node is Sandy Bridge.

This means that you have to do a funky dance because mpiexec is only on the login nodes but helper executables will explode in interesting ways if you try to make on the node. I'm experimenting with cmaking/configuring on the *login* node and making on a *compute* node.

The whole thing is really silly though.

gdsjaar:
Yes, welcome to the nightmare….  One possibility that I have used in the past is to get an interactive allocation of 1 or a few nodes and do the build/configure on that allocation.  You are on the compute node, so everything runs…. However, there are some compute nodes on some platforms that don’t have access to compilers, so that can cause issues.

We do have a contract with the HDF5 developer group and one of their tasks is to support cross-compilation on these types of systems, so it should get easier in the future.

..GReg

npe9:
Yup that's exactly the problem. My final (hopefully) issue with getting things up and running is that netcdf is compiling with pnetcdf but nc-config is saying that it's not enabled. This is causing Cmake to miss a bunch of symbols.

Is there anything wrong with this latest pull request that I should fix?

npe9:
More cuteness, some compute nodes can't go out on the network, so my build script now has the following:
```
INSTALL=NO ./install-tpl.sh
DOWNLOAD=NO srun ./install-tpl.sh
```

gsjaardema:
Yes, now you see why there is an INSTALL and DOWNLOAD.  Primarily for SCN machines, but also useful for others with no network access.

gsjaardema:
Will look at PR hopefully soon.  In middle of some meetings and other work...

gsjaardema:
Are you using current NetCDF, or an older version?  I thought we had fixed the nc-config bug, but maybe it has come back...

gsjaardema:
Thanks for the work.  The `install-tpl.sh` is fairly new and definitely needs some work.  One of the biggest issues is that it doesn't report problems very well -- just keeps going and up to user to decide where it went wrong and how to fix.

npe9:
@gsjaardema I'll see about addressing that in a separate PR.

PR #105, opened 2018-05-04
Major surgery on install-tpl.sh and some runcmake.sh changes for netcdf.

npe9:
I don't expect this one to work out of the box, so don't feel obligated to accept it. This PR is to make sure I'm on the right track more than anything.

I've thrown in some checks to avoid recompiling and some error checking to stop the script after a failed compile. I've also included a few more tweaks to make it easier to compile static libraries.


gsjaardema:
Looks reasonable and is an improvement over what is currently there.

Issue #91, opened 2017-12-20
You have a CMake bug


knepley:
Here is the fix (from 840b0e38a7390ecd53614f68b9b5def6b4a3fbd2),

diff --git a/cmake/tribits/common_tpls/find_modules/FindNetCDF.cmake b/cmake/tri
index a71d387..b88c4d0 100644
--- a/cmake/tribits/common_tpls/find_modules/FindNetCDF.cmake
+++ b/cmake/tribits/common_tpls/find_modules/FindNetCDF.cmake
@@ -175,7 +175,7 @@ else(NetCDF_LIBRARIES AND NetCDF_INCLUDE_DIRS)
         if(meta_path)
           # Search meta for NetCDF Version...
           file(STRINGS "${meta_path}/netcdf_meta.h" netcdf_ver_string REGEX "NC
-          string(REGEX REPLACE "[^0-9\.]" "" netcdf_version "${netcdf_ver_strin
+          string(REGEX REPLACE "[^0-9\\.]" "" netcdf_version "${netcdf_ver_stri
 
           # Search meta for NC_HAS_PARALLEL setting...
           file(STRINGS "${meta_path}/netcdf_meta.h" netcdf_par_string REGEX "NC

gsjaardema:
I seem to get the correct behavior for the current REGEX REPLACE.  Are you seeing a failure that caused you to suggest this patch?

knepley:
Yes

knepley:
Did you push the fix?

  Thanks,

    Matt

@jedbrown

gsjaardema:
I didn't push the fix since I haven't been able to reproduce the problem it is meant to solve and have had no other instances of anyone having the problem.  If you can show me the output indicating the problem and what system(s) it is appearing on, I will revisit.

PR #113, opened 2018-08-23
Fix configuration issue with netcdf < 4.5

bmpersc:
Issue #91

Fix a missing string that was used to determine the netcdf_max_vars. The missing
string caused that value to be empty.

Fixed the use of keyword "GREATER_EQUAL" in an if. That keyword didn't exist until
CMake 3.7 and the minimum version required is 3.1.


bmpersc:
@gsjaardema, The appveyor error is from Kokkos having files with invalid names on Windows. 

Kokkos_Experimental::ROCm_ViewCopyETIInst_int64_t_double_LayoutLeft_Rank1.cpp

There are two ":" in the name, which isn't valid in windows, I'm kinda surprised it was valid anywhere. However, this has nothing to do with the changes I made so I don't know what I can/should do about fixing it.

Issue #120, opened 2018-10-18
Segfault possible in aprepro_lib


bmpersc:
I recently ran across a use case where aprepro_lib segfaults in the Aprepro class destructor. The issue is the call to outputStream.top()->flush(). It appears in my use case that outputStream has no elements in it so this tries to call flush on an element that doesn't exist.

I have found a fix, which is just to check the size of outputStream is > 0 before calling the flush. I will issue a PR shortly.

PR #121, opened 2018-10-18
Aprepro_lib: fix potential segfault in destructor

bmpersc:
Issue #120

In certain usages of the Aprepro class the outputStream stack might
have no elements in it so calling flush on the top element segfaults.

The fix is to check that outputStream has > 0 elements.

### DIFF
--- a/packages/seacas/libraries/aprepro_lib/apr_aprepro.cc
+++ b/packages/seacas/libraries/aprepro_lib/apr_aprepro.cc
@@ -77,7 +77,7 @@ namespace SEAMS {
 
   Aprepro::~Aprepro()
   {
-    if ( outputStream.size() > 0) {
+    if (!outputStream.empty()) {
       outputStream.top()->flush();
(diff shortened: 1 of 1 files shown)


bmpersc:
@gsjaardema The appveyor issue is still the issue with Kokkos files having "::" in their names which isn't valid in windows. There is nothing I can do about that, the Kokkos developers will have to address it.

bmpersc:
@gsjaardema I'm fine with using empty instead of size. However, it looks like the change you made to use a ternary operator has caused the compile to fail.

I didn't know what hist.index was and I wasn't hitting that portion of the code in a bad way so I figured I'd leave well enough alone especially since that has been working for years.

gsjaardema:
Yes, I have submitted a fix to my fix.  It probably wouldn't hit the hist.index, but figured if I didn't change it there also, it would probably bite me at some future time...  Should build this time.

PR #134, opened 2019-01-07
Allow users to disable static library build

ibaned:
even with BUILD_SHARED_LIBS=ON, this repository
will compile a libexodus.a
This commit introduces an option SEACASExodus_ENABLE_STATIC
that can be set to OFF to prevent any static library
compilation.
It should not affect anyone who doesn't explicitly set this
new option.


ibaned:
Good point! I've pushed this additional change to the pull request.

PR #138, opened 2019-01-28
exodusII_int: use inttypes.h for VS2015 and newer

mathstuf:
`PRId64` is now provided by VS2015 and causes warnings about
redefinition if done here.


mathstuf:
The one CI failure is due to an `apt-get` error.

gsjaardema:
Thanks.

PR #140, opened 2019-03-26
Ioss can save exodus database using ADIOS2

fbudin69500:
The Adaptable Input/Output System (ADIOS) is A Unified I/O Performance
Framework [1]: for Transport and Transform of groups of self-describing data
variables across different media (file, wide-area-network, in-memory
staging, etc.) with performance and ease of use as the main goals.

Files written using the ADIOS2 library are saved using a flat structure. Each
IOSS variable is saved inside the ADIOS file as an independent ADIOS variable.
IOSS properties are saved in the same file as independent ADIOS variables.

ADIOS2 uses MPI out of the box. ADIOS2 files are step-based: Transient data
is natively handled by ADIOS2.

Using ADIOS2 IOSS database instead of the Exodus database should only require
the developer to specify "adios" instead of "exodus" when instanciating a
IOSS database object using the Ioss::IOFactory.

Current known limitations:
* A limited number of ADIOS2 parameters are accessible through the DatabaseIO
properties ("Engine_in", "Engine_out", "Transport", "Library", "IPAddrres",
"MarshalMethod", "verbose") [2].
* Limited streaming support with limited testing.
* HISTORY and HEARTBEAT format not supported.
* RESTART not supported yet.
* Limited testing.

Default settings:
* Engine=BPFile
* Transport=File
* Library=POSIX

[1] https://github.com/ornladios/ADIOS2
[2] https://adios2.readthedocs.io/en/latest/engines/engines.html


fbudin69500:
@chuckatkins First PR to incorporate ADIOS into IOSS.

fbudin69500:
Hello @gdsjaar,

This is a PR to add ADIOS2 as an IOSS database. This follows the conversation we had a couple months ago. Please let me know what you think should be improved.

gsjaardema:
@fbudin69500 Thanks.  I will try to take a look soon. Probably won't be until next week though.

cmake-devel
```diff
@@ -67,7 +67,7 @@ BUILD_TYPE="${BUILD_TYPE:-DEBUG}"
 ### which provides X11/Xlib.h and the libX11, then change the "ON"
 ### below to "OFF". It will disable blot and fastq
 HAVE_X11=${X11:-ON}
-HAVE_X11=OFF
+
```
fbudin69500: Thanks. I had hardcoded it locally to make the tests pass in a docker container, and I forgot to remove it afterwward.

fbudin69500:
@gsjaardema Thanks for the review and the corrections. I saw that some of the CI were read. I will take a look at that.

gsjaardema:
Let's see how the CI runs this time...

fbudin69500:
@gsjaardema Thanks so much for solving the final tests failures!

gsjaardema:
@fbudin69500 It looks like Travis uses a very old version of openmpi which was causing a build failure for `MPI=ON`.  I will have to solve that eventually so can use adios2 in parallel, but for now will just travis test it for serial.

I am getting a build error for not finding `H5Oget_info` when using an older version of HDF5 on linux.  Will let you know more details as I find them.  It looks like the HDF5 group is deprecating that function for performance reasons, but I'm not sure why it would be in a hdf5-1.10.5, but not in hdf5-1.10.3...

fbudin69500:
Sounds good. Let me know if you see anything that needs fixing on my side. If HDF5 is causing problems with ADIOS, I can deactivate it in ADIOS configuration as it is not necessary for CI.

gsjaardema:
The current adios2 TPL installation is checking out a specific commit sha, is this still needed, or can I use a more "standard" version of adios2:
```
git checkout 8c2135169ad534c30c503ad2bb8be0507facf63b
```

gsjaardema:
@chuckatkins  In case you have input on the above question regarding adios2 version

PR #142, opened 2019-04-25
Giving group read and execute perms for 'decomp' in binary dir (ATDV-156)

bartlettroscoe:
The strategy that I am using to install Trilinos (for SPARC testing and upgrades) using the 'jenkins' entity account requires that all of the files being installed be group readable so the install command can be run by another user in the group (the atdm-devops-admin user in our case, see [ATDV-156](https://sems-atlassian-srn.sandia.gov/browse/ATDV-156)). If the base build directory is properly protected with permissions this does not loosen the protection of this 'decomp' script at all (as it is just group readable not group writable).

This matches the PR trilinos/Trilinos#5008 and must be merged or the next snapshot from SEACAS to Trilinos will wipe out that patch.


bartlettroscoe:
Thanks @gsjaardema!

PR #155, opened 2019-11-29
style: replace import * with calling each thing

kayarre:
fixed a discrepancy with the versioning.

changed the ctype to be explicit,  ` from ctypes import *` to `import ctypes` which is better practice.

could also have only imported the relevant functions, but if enough are used to use `*` then using explicitly makes the most sense I think.


gsjaardema:
Thanks for the PR.

PR #166, opened 2020-04-22
Make install_executable fn check that target exists

brian-kelley:
Make install_executable(target) verify that target actually exists
before doing INSTALL. In Trilinos, executables added via
TRIBITS_ADD_EXECUTABLE won't be added if the build has different
CATEGORIES than the executable. If those mismatch, install_executable
caused an error because it would always try to install.


brian-kelley:
@gsjaardema See https://github.com/trilinos/Trilinos/issues/7232 for more discussion

brian-kelley:
Thanks, fixed.

gsjaardema:
Looks good now.  Thanks.

PR #176, opened 2020-06-08
Added two command line options to Aprepro. With --errors_fatal, no ou…

jbcarleton:
…tput

will be written if there are any errors. With --errors_and_warnings_fatal,
no output will be written if there are any errors or warnings.


gsjaardema:
Thanks.  Looks good.  Will run through clang-format once it is merged, but looks good to me.

PR #178, opened 2020-06-22
Fix MinGW build

jschueller:


jschueller:
@gsjaardema its ready now

gsjaardema:
thanks.

PR #191, opened 2020-10-14
Remove bad gitignore file

wortiz:
Removes a gitignore file which looks like it accidentally got copied over a long time ago, was causing some search tools which use gitignore to break


gsjaardema:
Thanks

PR #193, opened 2020-11-03
Add virtual destructor to SetItem

ibaned:
this is needed in order to call delete
on SetItem base class pointers, which
SEACAS does.


gsjaardema:
Thanks.

PR #197, opened 2020-11-05
APREPRO: Add --require_defined flag to treat undefined var as error

gsjaardema:
Pull over Sierra/SD generated PR to the seacas github repository.


gsjaardema:
@ibaned Here is the proposed change from Sierra/SD which is similar to your issue #197.  Let me know if you would like some changes.

ibaned (approved):
This is great! Our team would also like to have this.

PR #199, opened 2020-12-10
Increase size of magic array for null character

ibaned:
this prevents out of bounds access when the fifth
character is set to the null character for later
printing: the array needs to be five characters
long.


gsjaardema:
Thanks.  I know I fixed this last week, but can't see where I did it... I must not have pushed it yet and it is sitting in some open window I have somewhere...

ibaned:
That's happened to me a lot!

PR #200, opened 2020-12-10
fix globs for hash file installation

ibaned:
it looks like the files were simultaneously
moved out of the hash subdirectory and
globbed from the hash subdirectory for
installation.
this fixes the install glob to grab the
files from the src directory where they
are now.


gsjaardema:
Thanks.

Issue #202, opened 2020-12-16
TriBITS FindNetCDF out of date


ibaned:
I'm having issues compiling on Windows because the FindNetCDF CMake file executes commands that only work on Linux.

The `FindNetCDF.cmake` file in SEACAS is very different from the one you can find on TriBITS master, so I think the proper fix is to bring in the latest TPL files from TriBITS. Would you like me to open a PR for this?

ibaned:
based on #203 I'll try to fix the file in SEACAS instead

ibaned:
This was resolved for me by #204

PR #203, opened 2020-12-16
update FindTPL files from TriBITS

ibaned:
TriBITSPub/TriBITS@c1c110d


gsjaardema:
I think the correct task is to update the TriBITs versions with the SEACAS versions.  Mine are more current than what is currently in TriBITs.

ibaned:
Ah, okay! Well, my original issue #202 was that the latest `FindNetCDF.cmake` in SEACAS didn't work on Windows, so maybe I'll try to fix it on Windows and open a PR for that instead?

gsjaardema:
That would be better I think.

PR #204, opened 2020-12-18
Fixes for Windows

ibaned:
These are mainly fixes to the CMake TPL modules and also one fix for the order of Windows header includes. I've tested these fixes on Windows 10 with Visual Studio 2019 and CMake 3.19. I've also tested this branch on macOS.


gsjaardema:
Thanks.

PR #210, opened 2021-01-13
Fixes for static libraries and Windows

ibaned:
HDF5 calls their targets something based on both static/shared and component, so this makes it easier to find them by just letting HDF5 tell us what they are. This fixes the build for me when using all static libraries, because HDF5 is better able to include things like `zlib` in its exported targets.

### DIFF
--- a/cmake/tribits/common_tpls/find_modules/FindHDF5.cmake
+++ b/cmake/tribits/common_tpls/find_modules/FindHDF5.cmake
@@ -373,7 +373,7 @@ else()
       NO_MODULE)
 
     # If located a HDF5 configuration file
-    if (HDF5_FOUND)
+    if (HDF5_CONFIG)
 
       message(STATUS "Found CMake configuration file HDF5 ( directory ${HDF5_ROOT} )")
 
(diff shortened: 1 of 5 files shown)


ibaned:
I don't think these test failures can be caused by a build system change. @gsjaardema can we re-run the tests?

gsjaardema:
All of the failures were using MPI=ON, so that is concerning, but I will restart to see if there was just some glitch with MPI that day.  If it fails again, I will try locally to see if it is repeatable...

ibaned:
Added some more fixes as I'm attempting to build on Windows with MPI

gsjaardema:
Error seems to be:
```
usr/bin/ld: cannot find -lhdf5-static
/usr/bin/ld: cannot find -lhdf5-shared
/usr/bin/ld: cannot find -lhdf5_tools-static
/usr/bin/ld: cannot find -lhdf5_tools-shared
/usr/bin/ld: cannot find -lhdf5_hl-static
/usr/bin/ld: cannot find -lhdf5_hl-shared
collect2: error: ld returned 1 exit status
packages/seacas/libraries/exodus/CMakeFiles/exodus.dir/build.make:4420: recipe for target 'packages/seacas/libraries/exodus/libexodus.so.2.0' failed
make[2]: *** [packages/seacas/libraries/exodus/libexodus.so.2.0] Error 1
CMakeFiles/Makefile2:1634: recipe for target 'packages/seacas/libraries/exodus/CMakeFiles/exodus.dir/all' failed
```
I'm not sure why non-mpi builds work but mpi builds fail...

gsjaardema:
Looks like it is working now.  Thanks for the hard work on this...

ibaned:
Thank you for being a very responsive maintainer!

PR #211, opened 2021-01-15
EXODUS: ex_get_name now works with assemblies.

mvlopri:
Use the ex_get_assembly function to get the name of assemblies
in ex_get_name for consitency across the data structures.


gsjaardema:
Thanks.

PR #218, opened 2021-02-26
remove space at start of shebang

wortiz:
I don't use a bourne style shell so this was breaking the script for me


gsjaardema:
Thanks.

PR #222, opened 2021-03-30
IOSS: See if we can eliminate the need for copy_database_v2

gsjaardema:
The only current difference between the `copy_database` and `copy_database_v2` functions were:
* use of `options.define_geometry`
* does not support `options.boundary_sideset`

This modifies `copy_database_v2` to just  set `options.boundary_sideset = false` and also adds the use of `options.define_geometry` to the `copy_database` function.  This option defaults to `true`

After (or as part of this), we can see if the new way of handling the `boundary` faces for the define_boundary_sideset case will work for Catalyst, or if further restructuring is needed.


gsjaardema:
@tjotaha Can you take a look at this and see if it works correctly for Catalyst.  Also see what Jeff thinks.
Thanks.

jeffmauldin:
I think we should also add a test for if the user chooses boundary_sideset=true and define_geometry=false, because this pathway causes a coredump. We could either test for that state and issue an error or, more simply, when we check if "boundary_sideset" is true, we "and" it with "define_geometry" so that we only try to do boundary_sidesets if we also define_geometry.

I am doing a SEACAS build right now off this branch and I will be building the plugin to make sure it works, but I am confident this change is good. I will comment again after I test.

jeffmauldin:
/gsjarrdema /tjotaha
I have now tried this branch with the changes to copy_database, and it all works perfectly. There is no need for the extra test described in my previous comment, as the code as written does not follow a bad path when the user sets boundary_sideset=true and define_geometry=false.

In addition to making sure this worked, I changed the call to copy_database_v2() in our test code to copy_database(), and that now works correctly. So we can remove the copy_database_v2() method as long as we change "copy_database_v2" to "copy_database" in the single file:
.../ioss/src/visualization/catalyst/ioss2catalyst/IossApplication.cxx
(there is only one instance of this method and it is on line 797)

gsjaardema:
Thanks, I will make those changes and then merge the pull request.

Thanks for prompting me to take a look at `copy_database` and "modularize" it a little.  I think there are some additional cleanups that can be made, but for now it is a little cleaner and we can eliminate some duplicated code.

gsjaardema:
NOTE: Only real change to `IossApplication.cxx` is changing `copy_database_v2` to `copy_database`.  It got run through the clang-format process, so it looks like there are lots of changes...

PR #223, opened 2021-04-16
exodus.py: Provide exodus.py with the get_assemblies function

mvlopri:
Allow for the ex_get_assemblies function to be called from
the exodus python interface. Includes a unittest as well.
Updated some Copyrights.

### DIFF
--- a/packages/seacas/libraries/exodus/src/ex_get_assemblies.c
+++ b/packages/seacas/libraries/exodus/src/ex_get_assemblies.c
@@ -1,5 +1,5 @@
 /*
- * Copyright(C) 1999-2020 National Technology & Engineering Solutions
+ * Copyright(C) 1999-2021 National Technology & Engineering Solutions
  * of Sandia, LLC (NTESS).  Under the terms of Contract DE-NA0003525 with
  * NTESS, the U.S. Government retains certain rights in this software.
  *
--- a/packages/seacas/scripts/exodus3.in.py
+++ b/packages/seacas/scripts/exodus3.in.py
@@ -1,5 +1,5 @@
 """
-exodus.py v 1.20.2 (seacas-beta) is a python wrapper of some of the exodus library
+exodus.py v 1.20.3 (seacas-beta) is a python wrapper of some of the exodus library
 (Python 3 Version)
 
 Exodus is a common database for multiple application codes (mesh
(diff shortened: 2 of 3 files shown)


packages/seacas/libraries/exodus/src/ex_get_assemblies.c
```diff
@@ -17,6 +17,8 @@
  */
 int ex_get_assemblies(int exoid, ex_assembly *assembly)
 {
+  EX_FUNC_ENTER();
```
gsjaardema, line 20: These aren't allowed here since `ex_get_assembly` has a call to `EX_FUNC_ENTER()`.  This macro is used for the thread-safe build and locks a mutex, so if this call is here and also in `ex_get_assembly()`, it will deadlock.

packages/seacas/libraries/exodus/src/ex_get_assemblies.c
```diff
@@ -26,7 +28,7 @@ int ex_get_assemblies(int exoid, ex_assembly *assembly)
     snprintf(errmsg, MAX_ERR_LENGTH, "ERROR: failed to inquire ASSEMBLY count in file id %d",
              exoid);
     ex_err_fn(exoid, __func__, errmsg, num_assembly);
-    return (EX_FATAL);
+      EX_FUNC_LEAVE(EX_FATAL);
```
gsjaardema, line 31: Same

packages/seacas/libraries/exodus/src/ex_get_assemblies.c
```diff
@@ -53,8 +55,8 @@ int ex_get_assemblies(int exoid, ex_assembly *assembly)
 
     int status = ex_get_assembly(exoid, &assembly[i]);
     if (status != EX_NOERR) {
-      return status;
+      EX_FUNC_LEAVE(status);
```
gsjaardema, line 58: same

packages/seacas/libraries/exodus/src/ex_get_assemblies.c
```diff
@@ -53,8 +55,8 @@ int ex_get_assemblies(int exoid, ex_assembly *assembly)
...
     }
   }
-  return EX_NOERR;
+  EX_FUNC_LEAVE(EX_NOERR);
```
gsjaardema, line 61: same

gsjaardema (approved):
Other than the `EX_FUNC_ENTER() / LEAVE()` calls, it looks good to me.  I know the names of these are not very descriptive and look more llike trace macros, but they are for locking a global mutex so cannot call any function which also locks the same mutex.

gsjaardema (approved):

PR #229, opened 2021-05-14
Reformulation of PR228

gsjaardema:
@tjotaha
Catalyst include file path fix …

Adjusted include file paths in Catalyst plugin build to
match IOSS include file convention. Updated CMakeLists.txt in
visualization and catalyst directories to remove uneeded
subdirectories.

Added fix to Phactori tests for seg fault with Python unittest
and pvbatch.

Included fix in Iovs_Utils.C and Iovs_Utils.h for unused boolean
parameter in method getCatalystPluginPath().

The original #228 was baselined against the incorrect repository I think.  Redid it against the current repository state to clean it up.  No changes to any of the visualization files from original PR


gsjaardema:
@tjotaha I redid the PR to clean it up somewhat.  Not sure what was wrong with original one, but it was showing diffs of several thousand unrelated files.

PR #230, opened 2021-06-01
Fix win32 build

jschueller:


packages/seacas/libraries/ioss/src/cgns/Iocgns_Utils.h
```diff
@@ -66,8 +66,8 @@ namespace Iocgns {
     Utils()  = default;
     ~Utils() = default;
 
-    static const size_t CG_CELL_CENTER_FIELD_ID = 1ul << 33;
-    static const size_t CG_VERTEX_FIELD_ID      = 1ul << 34;
+    static const size_t CG_CELL_CENTER_FIELD_ID = 1ul << 30;
```
gsjaardema, line 69: This will probably work OK, since I doubt that there will be on order 2^30 fields ever used, but the intent was to leave the lower 32-bits untouched.  This should work, but I will also add a test to make sure that the incoming field index is < 2^30 so that the bit doesnt' get mixed in with the index.  I guess I should have had a static assert that sizeof(size_t) > 32 with the olde code...  Thanks for finding this.

gsjaardema (approved):

packages/seacas/libraries/ioss/src/cgns/Iocgns_Utils.h
jschueller: yes, the problem is that size_t is 4 bytes wide on win32

PR #241, opened 2021-07-30
Capture arrays by reference rather than value

cquammen:
MSVC 2015 doesn't compile when capturing stack-allocated arrays by
value in a lambda expression.


gsjaardema:
Thanks.

PR #242, opened 2021-07-30
Add Catalyst 2.0 support as a Database to IOSS

utkarshayachit:
This is a WIP MR to add support for Catalyst 2.0 as a Database to IOSS.


gsjaardema (commented):
On a quick first look it looks reasonable.  There is definitely some work to be done, but should be able to demonstrate minimal capability which is the intent I think.  I don't have any complaints on the syntax or implementation.  

I will try to build later this week hopefully

utkarshayachit:
> @utkarshayachit I built the code with the PR, but when I run a `io_shell input.exo output.exo`, I get an exception failure. It looks like Catalyst2 is throwing an exception in the Catalyst `show_configuration()` routine. I'm not sure if some initialization or what is missing. Note that in this case, there is no catalyst input or output file, but it is just trying to print the configuration of the library. I get the same failure with a simple `io_info --show-config`

@gsjaardema, this should be fixed now. I also cleaned up the branch by rebasing on `master` and cleaning up any conflicts.

utkarshayachit:
thanks for the merge! I'll have VTK/ParaView updated next.

PR #243, opened 2021-08-09
Ioss: Missing strncasecmp define

jschueller:
seems necessary for some msvc version


mathstuf:
At least VS2019 toolchains.

gsjaardema:
Thanks.

PR #245, opened 2021-08-17
gcc 4.8.5: use non-const iterators for std::vector::erase()

cquammen:
Workaround gcc 4.8.5's lack of support for const iterators
as arguments to std::vector::erase().


gsjaardema:
Thanks.  I actually just hit this issue yesterday with the IBM compiler...

PR #258, opened 2021-11-02
Ioex_IOFactory: remove zlib.h usage

mathstuf:
This isn't actually used anywhere else in ioss, so its version isn't
very useful.

### DIFF
--- a/packages/seacas/libraries/ioss/src/exodus/Ioex_IOFactory.C
+++ b/packages/seacas/libraries/ioss/src/exodus/Ioex_IOFactory.C
@@ -16,7 +16,6 @@
 #include <exodusII.h>
 #include <fmt/ostream.h>
 #include <string> // for string
-#include <zlib.h>
 
 #include "Ioss_CodeTypes.h" // for MPI_Comm
 #include "Ioss_DBUsage.h"   // for DatabaseUsage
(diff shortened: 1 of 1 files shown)


gsjaardema:
The usefulness to me and why I included it is that we are testing the zlib-ng which is much faster than normal zlib compression.  It is difficult to determine whether someone is using that zlib library or not and this provided a way to track this and help in debugging their builds or performance issues they had during read/write of compressed data.

mathstuf:
But where else is `zlib.h` included? I only see it for this information, so where is it getting used?

mathstuf:
FWIW, there's no zlib logic in the `CMakeLists.txt` files, so what `zlib.h` this is getting may have no relation to the one used by HDF5 or NetCDF (where it seems to be used).

gsjaardema:
It is included in the HDF5 and netCDF libraries.  The include path is specified during the build process, so *if* the build is a TPL / SEACAS build and doesn't use system-installed or externally-installed libraries, then the zlib should be the same.  Not perfect, but it is working for what I need.  I can probably take a look and see if there is a more robust method, but that would still require calling a zlib function which would then require still including the zlib.h.

There have been issues in the past where it is difficult to determine what libraries and builds a user was using at the time of the exodus file generation or also debugging execution issues of someone's seacas build, so whatever information I can access from an application or a database, the easier it can be to debug problems.

mathstuf:
Hmm. OK. I've wired it up to `vtk_zlib.h` in VTK already, but this was to see if removal was eventually better.

PR #262, opened 2021-11-09
Text mesh

tokusanya:
Enable the use of STK capability, TextMesh in IOSS as a means of quickly generating test meshes especially in the context of unit testing. I've been working with Greg on this and gotten his feedback on Sandia gerrit. I think it's in a decent state and have also included a unit test file to make sure that it does not diverge from the STK implementation


tokusanya:

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMesh.h
```diff
@@ -0,0 +1,366 @@
...
+    };
+  }
+};
+
+using Topology = TopologyMapEntry;
+using TextMeshData = text_mesh::TextMeshData<int64_t, TopologyMapEntry>;
+using ElementData = text_mesh::ElementData<int64_t, TopologyMapEntry>;
+using Coordinates = text_mesh::Coordinates<int64_t, TopologyMapEntry>;
+using TextMeshParser = text_mesh::TextMeshParser<int64_t, IossTopologyMapping>;
+using MapVector = std::vector<int64_t>;
```
gsjaardema, line 160: This is the same as an `Ioss::Int64Vector`.  Since we use the `Ioss::IntVector` below, we should probably be consistent and use `Ioss::Int64Vector` instead of `MapVector` unless there is something I am overlooking.  Was just confusing when I was looking at functions below and wanting to know the difference between them...

This probably came from the Iogn_GeneratedMesh originally which should also be updated...

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMesh.C
```diff
@@ -0,0 +1,626 @@
...
+{
+  if (m_rawCoordinates.empty()) return;
+
+  /* create global coordinates */
+  int64_t count = node_count_proc();
+  x.reserve(count);
+  y.reserve(count);
+  z.reserve(count);
+
+  auto nodes = m_data.nodes_on_proc(m_myProcessor);
```
gsjaardema, line 410: Should this be a reference or const reference?

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMesh.h
```diff
@@ -0,0 +1,366 @@
...
+  std::set<int64_t> get_local_element_ids_for_block(int64_t id) const;
+
+  unsigned m_dimension{3};
+  size_t m_processorCount{0};
+  size_t m_myProcessor{0};
+
+  size_t m_timestepCount{0};
+  std::map<Ioss::EntityType, size_t> m_variableCount;
+
+  std::vector<double> m_rawCoordinates;
```
gsjaardema, line 354: It looks like this is populated during parsing.  It is then used to fill the vectors in `m_coordinates` at which time we have the same/similar data in both locations.  After filling, it looks like the only use of `m_rawCoordinates` is to check whether it is empty.  I realize that text meshes are usually "small", but seems like a waste to keep similar data in two locations.  Maybe I am missing the need for both.

gsjaardema (changes requested):
Looks very good.  I had a few comments and questions about some possible modifications.  Not sure if any are required.  

Need to determine what testing framework to use.  I have been using Catch2 and I see that your unit test uses google test.  If Catch2 doesn't support what is needed, then we can probably add google test to the list of TPLs...

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMesh.h
```diff
@@ -0,0 +1,366 @@
...
+#include <Ioss_CodeTypes.h>
+#include <Ioss_EntityType.h>  // for EntityType
+
+#include <array>
+#include <cstddef>  // for size_t
+#include <cstdint>  // for int64_t
+#include <map>      // for map, etc
+#include <string>   // for string
+#include <utility>  // for pair
+#include <vector>   // for vector
```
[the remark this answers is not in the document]
tokusanya: This is true .. I can make that change

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMesh.C
```diff
@@ -0,0 +1,626 @@
...
+  m_variableCount[Ioss::REGION] = 0;
+
+}
+
+void TextMesh::parse_coordinates_option(const std::vector<std::string> &option)
+{
+  if (option.size() > 1) {
+    auto tokens = Ioss::tokenize(option[1], ",");
+    m_rawCoordinates.reserve(tokens.size());
+    for (auto token : tokens) {
```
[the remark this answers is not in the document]
tokusanya: Sure, easy fix. For the most part, Timothy had insisted that TextMesh would only be used for small meshes so that performance was not of paramount concern .... but I'm sure there is that one user out there that will abuse this ;)

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMesh.C
```diff
@@ -0,0 +1,626 @@
...
+{
+  /* create global coordinates */
+  size_t count = node_count_proc();
+  xyz.resize(count);
+  coordinates(component, xyz.data());
+}
+
+void TextMesh::coordinates(int component, double *xyz) const
+{
+  auto nodes = m_data.nodes_on_proc(m_myProcessor);
```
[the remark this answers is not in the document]
tokusanya: The nodes_on_proc() function actually returns a std::set object and not a reference. It does a search for the processor argument on an unordered map and if it can't find it, it returns an empty set. I suppose the TextMeshData struct could hold an empty set and return that if it could not find the processor argument....this would allow a reference to be returned. Let me see what can be done about that

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMesh.C
```diff
@@ -0,0 +1,626 @@
...
+TextMesh::TextMesh(const std::string &parameters, int proc_count, int my_proc)
+    : m_processorCount(proc_count), m_myProcessor(my_proc)
+{
+  // Possible that the 'parameters' has the working directory path
+  // prepended to the parameter list.  Strip off everything in front
+  // of the last '/' (if any)...
+  auto params = Ioss::tokenize(parameters, "/");
+
+  if (!params.empty()) {
+    auto groups = Ioss::tokenize(params.back(), "|+");
```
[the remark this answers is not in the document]
tokusanya: This is an unfortunate side effect of having to manually extend the current implementation of TextMesh in STK. The base implementation was only set up to parse topology and connectivity, and then an option is given to the user to apply a set of coordinates to the parsed data. In order to extend this functionality to a string based mesh description, I had to essentially parse and strip out the coordinate data from the string and then have the base implementation do the parsing of topology/connectivity, after which I pass along the coordinate data. You are right that extension of the TextMesh language to include sidesets and other groupings will run into the same issue. Maybe the thing to do is to get the STK team (Timothy specifically) to formalize the syntax and have all the parsing done within the base implementation

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMesh.C
```diff
@@ -0,0 +1,626 @@
...
+  initialize();
+}
+
+TextMesh::TextMesh(const std::string &parameters, int proc_count, int my_proc)
+    : m_processorCount(proc_count), m_myProcessor(my_proc)
+{
+  // Possible that the 'parameters' has the working directory path
+  // prepended to the parameter list.  Strip off everything in front
+  // of the last '/' (if any)...
+  auto params = Ioss::tokenize(parameters, "/");
```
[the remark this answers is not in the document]
tokusanya: Oops, this was a copy/paste from StkMeshIoBroker ... I can remove this

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMesh.h
tokusanya: Yup, this was from Iogn_GeneratedMesh ... I'll remove the MapVector definition

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMesh.C
tokusanya: Refer to previous comment

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMesh.h
tokusanya: This is correct ... I was actually thinking of optimizing storage by not keeping it around longer that necessary. Let me see what I can do about making it a temporary object and I'll upload a new commit

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMesh.C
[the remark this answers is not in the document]
gsjaardema: OK, that sounds good.

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMesh.h
gsjaardema: I will try to fix elsewhere...

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMesh.h
tokusanya: A thought occured to me that if we were to formalize the syntax of TextMesh such that the parsing of coordinates/sidesets/nodesets/etc was done natively in TextMesh, then there wouldn't be a need for this container for the raw coordinates.

gsjaardema: Yes, that sounds right.  I think that would be good...

gsjaardema (approved):

gsjaardema:
Thanks.

PR #266, opened 2021-11-19
Ioss_CodeTypes: avoid defining ssize_t

mathstuf:
Instead, make an Ioss-specific typename to avoid conflicting with other
libraries using `ssize_t` in their code (either consumers of Ioss or
third parties that Ioss itself uses).

### DIFF
--- a/packages/seacas/libraries/ioss/src/Ioss_CodeTypes.h
+++ b/packages/seacas/libraries/ioss/src/Ioss_CodeTypes.h
@@ -20,10 +20,13 @@
 
 #if defined(__IOSS_WINDOWS__)
 #ifdef _WIN64
-#define ssize_t __int64
+#define ioss_ssize_t __int64
 #else
-#define ssize_t long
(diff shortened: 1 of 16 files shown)


gsjaardema:
Thank you for the contribution.  I've looked through the uses of `ssize_t` and found:
* Some are unused (a few includes)
* Some could be `size_t`.  
* Some could just be an `int`
* The rest could just as easily be `int64_t` and eliminate a define and windows-specific ifdefs.

I've gone through the code and eliminated all uses of `ssize_t`.  

This PR is welcome, but I think eliminating the use altogether might be a better strategy...

mathstuf:
Removal is also fine :) . I'm more interested in getting rid of the `#define ssize_t` than anything else.

gsjaardema:
OK, my changes have  been comitted. Should not be any `ssize_t` anywhere in the source code.  I will close this PR.  Thanks for  prodding me to take a closer look at the `ssize_t` uses...

PR #268, opened 2021-12-02
IOSS: Unify TextMesh unit tests in Ioss and STK

tokusanya:
1) Created single unit test file that can be copied between systems
2) Created associated test fixtures that support the test


gsjaardema (approved):
Looks good.  Will there be a later commit to actually run the tests via ctest?

Minor issue -- copyright range should end on 2021.

tokusanya:
> Looks good. Will there be a later commit to actually run the tests via ctest?
> 
> Minor issue -- copyright range should end on 2021.

I was looking high and low for how SEACAS runs gtest unit tests but could not find anything .. was that removed? If so, how do I enable the tests in ctest?

gsjaardema:
There currently aren't any gtest tests in SEACAS except for if the ADIOS2 library is built and in that case, only the Adios2 TPL uses gtest.  

IOSS testing currently uses [catch2](https://github.com/catchorg/Catch2).  

We could either:

- Add gtest TPL download and install and then enable the test if `TPL_ENABLE_GTest` is defined, or
- Change the test to use catch2.

I'm fine with either one, it probably makes sense to have gtest as an option and the tpl_install.sh is already set up to download it and install it, so you would just have to test whether `TPL_ENABLE_GTest` is enabled and if so, build the test executable and then run the test.

tokusanya:
Ok, let me look into how ADIOS2 uses gtest and maybe I can figure something out

PR #271, opened 2021-12-13
Activate text mesh unit tests

tokusanya:
The changes are based on the conversation I'm having with Greg. Let me know if there are issues with the naming conventions especially if there is some possibility that there might be other future unit tests not related to TextMesh


gsjaardema:
Looks good...

PR #275, opened 2022-01-17
ioss: avoid conflicting with MPI type definitions

mathstuf:
Instead of defining MPI types directly, map into an Ioss-specific name
to avoid conflicts.


gsjaardema:
I'm still trying to decide on this one.  In the meantime, I did eliminate the potential redefinition of MPI_COMM_{WORLD, SELF, NULL} in Ioss_CodeTypes.h and isolated it in Ioss_ParallelUtils.h...

mathstuf:
That is better, but it probably needs even more hiding if that header is included by Ioss users (even transitively).

gsjaardema:
At this point, the only change is `MPI_Comm` to `Ioss_MPI_Comm`.

I need to look at the use cases to see what this means for the various combinations:

- IOSS parallel-enabled, client code parallel-enabled --> OK
- IOSS not-parallel-enabled, client code not-parallel-enabled --> OK
- IOSS parallel-enabled, client code not-parallel-enabled --> Different definition for Ioss_MPI_Comm; is that OK?
- IOSS not-parallel-enabled, client code parallel-enabled --> Different definition for Ioss_MPI_Comm; is that OK?

This does eliminate the issue that IOSS would be redefining `MPI_Comm` for the client code in the last two cases, but I'm not sure whether it eliminates issues of the client code actually using a differently-compiled IOSS (and exodus) library.

mathstuf:
It should really be based on "is IOSS MPI-aware", not "is this current compilation MPI-aware". A header getting a `#define IOSS_KNOWS_ABOUT_MPI 0/1` line would make it a lot more reliable to decide what `Ioss_MPI_Comm` should be.

gsjaardema:
And that is what `SEACAS_HAVE_MPI` currently does.

I guess I'm still worried that if SEACAS/IOSS is compiled with `SEACAS_HAVE_MPI` undefined and the client is compiled with parallel-enabled, then if they pass in `MPI_COMM_WORLD` to the IOSS library, they may expect parallel behavior, but they won't get it and they won't get an error or warning...   The `MPI_COMM_WORLD` will be converted to an integer...

Alternatively, if IOSS is compiled with parallel enabled (`SEACAS_HAVE_MPI` defined), but the client is a serial non-mpi build, then `MPI_Init` will never be called and on some systems this can cause issues.  There will also likely be missing libraries that the mpi-enabled IOSS code requires, but the non-mpi client code will not add to the link line...

PR #277, opened 2022-01-17
IOSS: Extended TextMesh syntax to sidesets and nodesets

tokusanya:


tokusanya:
@gdsjaar Hope you got my email about this

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
```diff
@@ -14,593 +14,1754 @@
...
+struct SideBlockInfo {
+  std::string name;
+  std::string sideTopology;
+  std::string elementTopology;
+  std::string touchingBlock;
+  std::vector<size_t> sideIndex;
+  unsigned numNodesPerSide;
+};
+
+enum SplitType { SPLIT_BY_TOPOLOGIES = 1, SPLIT_BY_ELEMENT_BLOCK = 2, SPLIT_BY_DONT_SPLIT = 3, SPLIT_INVALID = 4 };
```
gsjaardema, line 362: Should get this from Ioss_SurfaceSplit.h instead of recreating it here.

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
```diff
@@ -14,593 +14,1754 @@
...
+
+inline std::ostream &operator<<(std::ostream &out, const SplitType &t)
```
gsjaardema, line 364: If this is needed, should be added to Ioss_SurfaceSplit.h for consistency and robustness.  Two definitions in two places will probably lead to then getting out of sync.

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
```diff
@@ -14,593 +14,1754 @@
...
+inline std::string strip(const std::string &inpt)
+{
+  auto start_it = inpt.begin();
+  auto end_it = inpt.rbegin();
+  while (std::isspace(*start_it)) ++start_it;
+  while (std::isspace(*end_it)) ++end_it;
+  return std::string(start_it, end_it.base());
+}
+
+inline std::vector<std::string> tokenize(const std::string &str, const std::string &separators)
```
gsjaardema, line 52: Would be better to use the existing IOSS tokenize function in `tokenize.h` to avoid duplicate code.

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
```diff
@@ -14,593 +14,1754 @@
...
-      handle_error<std::logic_error>(message);
+    if (second == std::end(str)) {
+      break;
+    }
+    first = std::next(second);
+  }
+  return tokens;
+}
+
+inline void convert_to_upper_case(std::string &str)
```
gsjaardema, line 70: Is it possible to use the Ioss::Utils::uppercase function.

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
```diff
@@ -14,593 +14,1754 @@
...
+{
+  std::transform(str.begin(), str.end(), str.begin(), ::toupper);
+}
+
+inline void convert_to_lower_case(std::string &str)
```
gsjaardema, line 75: Is it possible to use the Ioss::Utils::lowercase function

packages/seacas/libraries/ioss/src/unit_tests/UnitTestIotmTextMeshFixture.h
```diff
@@ -170,6 +165,85 @@ namespace Iotm {
...
+        size_t count = m_region->get_sidesets().size();
+        EXPECT_EQ(goldCount, count);
+      }
+
+      void verify_sideset_subset(const Ioss::SideSet* sideset, const unsigned id, const std::vector<std::string>& subsetNames)
+      {
+        EXPECT_TRUE(nullptr != sideset);
+        EXPECT_EQ(id, sideset->get_property("id").get_int());
+
+        if(subsetNames.size() == 0) {
```
gsjaardema, line 180: `subsetNames.empty()` preferred

gsjaardema (commented):
Looks good.  Seems to be some code duplicated with existing Ioss code.  Maybe a good reason for the duplication, but if not would be good to use the existing code.

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
tokusanya: Please see above. The options are definitely related to Ioss but don't necessarily have to be the same

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
tokusanya: Please see the first comment on independent portability. The tokenize function was taken from tokenize.h and modified slightly for TextMesh use

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
tokusanya: Please see first comment on independent portability

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
tokusanya: Please see first comment on independent portability

packages/seacas/libraries/ioss/src/unit_tests/UnitTestIotmTextMeshFixture.h
tokusanya: I can fix that

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
gsjaardema: I guess I don't understand why there is a need to have TextMesh be independent of the system.  Why is  a pure STK implementation needed and a pure IOSS implementation and a pure ??? implementation?  It makes the code confusing and duplicates code, but I don't see the benefit.   

The only potential benefit I see is that you can possibly test STK independent of IOSS, but if you are simply replicating code I'm not sure that is a benefit...  In other words, is there truly a need for an independent "SplitType" since that is purely an IOSS construct.  If I grep `SplitType` in the STK code, the only returns are `Ioss::SurfaceSplitType`

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
gsjaardema: If this is truly needed, then we should remove the `#include <tokenize.h>` in Iotm_TextMeshUtils.h and Iotm_TextMesh.C and probably rename it to make it very clear which tokenize is being used.

tokusanya: Good point

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
tokusanya: Yes, the design was to be able to test STK independent of IOSS because we have multiple builds of STK with and without STK_IO. That's one reason TextMesh was set up to be independent of what system it was being used in. The version I mentioned earlier is the STK Mesh concrete implementation of TextMesh which does not go through STK_IO. The TextMesh sideset split code has not been pushed into STK Mesh yet which is why your grep didn't show it. I was working on getting the STK and IOSS versions ready to go at the same time

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
```diff
@@ -14,593 +14,1778 @@
...
 
-    template <typename T> class TopologyMapping
-    {
-    public:
-      using Topology = T;
+    return names;
+  }
 
-      virtual ~TopologyMapping() {}
+  std::vector<std::string> get_part_names() const { return m_partNames; }
```
gsjaardema, line 210: Should this be `const std::vector<std::string> &get_part_names()`  It looks like the returned vector is never modified, so a const reference would be more efficient.

gsjaardema (approved):
Looks much better.  Comments are mainly me just making sure I understand the code.

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
```diff
@@ -410,7 +428,9 @@ class EntityGroup
     set_error_handler([](const std::ostringstream &errmsg) { default_error_handler(errmsg); });
   }
 
-  void set_error_handler(ErrorHandler errorHandler) { m_errorHandler = errorHandler; }
+  virtual ~EntityGroup() {}
```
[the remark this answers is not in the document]
tokusanya: My thought was to make sure that it wasn't default constructable which is why the default constructor is in the protected section.  But yes, using default and delete probably works better

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
```diff
@@ -1764,4 +1787,4 @@ class TextMeshParser
 };
 
 }  // namespace text_mesh
-}  // namespace Iotm
+}  // mamespace Iotm
```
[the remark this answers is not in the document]
tokusanya: hehe, typo

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
```diff
@@ -704,14 +723,14 @@ class SidesetSplitter
...
-  void create_index_proc_map(
+  void build_index_proc_map(
       const SidesetData<EntityId, Topology> &sideset, const std::vector<ElementData<EntityId, Topology>> &elementData)
   {
     for (size_t i = 0; i < sideset.data.size(); ++i) {
       const SidesetDataType<EntityId> &elemSidePair = sideset.data[i];
       EntityId elemId = elemSidePair.first;
 
-      auto iter = std::find(elementData.begin(), elementData.end(), elemId);
+      auto iter = bound_search(elementData.begin(), elementData.end(), elemId);
```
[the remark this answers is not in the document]
gsjaardema: OK.  That makes sense.  Is it ok that the sort uses a custom comparator, but the bound_search just uses the default less than comparator?

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
```diff
@@ -1510,8 +1532,9 @@ class TextMeshOptionParser
         "element list .. first "
         "argument)\n"
         "\tcoordinates:x_1,y_1[,z_1], x_2,y_2[,z_2], ...., x_n,y_n[,z_n] (specifies coordinate data)\n"
-        "\tsideset:[name,]elem_1,side_1, elem_2,side_2, ...., elem_n,side_n (specifies sideset data)\n"
-        "\tnodeset:[name,]node_1, node_2, ...., node_n (specifies nodeset data)\n"
+        "\tsideset:[name=<name>;] data=elem_1,side_1,elem_2,side_2,....,elem_n,side_n; [split=<block|topology|none>;] "
+        "(specifies sideset data)\n"
+        "\tnodeset:[name=<name>;] data=node_1,node_2,....,node_n (specifies nodeset data)\n"
         "\tdimension:spatialDimension (specifies spatial dimension .. default is 3)\n"
         "\thelp -- show this list\n\n");
```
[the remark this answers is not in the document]
tokusanya: Hmm, let me test that ...

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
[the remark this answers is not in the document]
gsjaardema: The delete makes it clearer that it isn't default constructable.

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
```diff
@@ -14,593 +14,1778 @@
...
+    return get_part_name(id);
+  }
+
+  unsigned size() const
+  {
+    if (!m_idsAssigned) assign_ids();
+    return m_ids.size();
+  }
+
+  std::vector<std::string> get_part_names_sorted_by_id() const
```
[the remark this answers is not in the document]
tokusanya: It's as a result of m_parts being a map so the extracted vector is guaranteed to be sorted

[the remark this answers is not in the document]
gsjaardema: Looks like m_parts is an `std::unordered_map` which isn't sorted...

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
tokusanya: Yep, you're right

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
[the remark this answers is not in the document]
tokusanya: Ok, I have fixed the text string for the help message. You can trigger the help message in the same way that you would with generated mesh via

`io_shell  -in_type textmesh "0,1,HEX_8,1,2,3,4,5,6,7,8|help" tmp.g`

It's based on the same parsing as GeneratedMesh and should obey the same rules.

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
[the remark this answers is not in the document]
tokusanya: That ..... is an interesting question. For the specific implementation of the custom comparator, it should be the same result as the default less because it is converting the ElementData entries into a simple EntityId. However, I see your point about consistency

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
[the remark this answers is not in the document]
tokusanya: `mutable std::map<unsigned, std::string> m_parts;`

It's m_ids that is a std::unsorted_map

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
tokusanya: This is fixed in the latest commit

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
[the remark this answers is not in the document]
gsjaardema: You're right.  I swear when I looked at it earlier they were both unordered_maps...  Sorry for the confusion.

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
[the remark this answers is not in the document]
tokusanya: I'm adding an ElementDataLess comparator that can be used in the sort and search. Will push after I test it

gsjaardema:
Thanks.  Looks good.

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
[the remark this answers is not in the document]
gsjaardema: Thanks.  It probably makes sense to see if just using "help" can be made to show the help information (That is something I have meant to do with GeneratedMesh for awhile too...)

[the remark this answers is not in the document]
tokusanya: Sounds like an opportunity to have it fixed up in both places :)

PR #279, opened 2022-02-07
EXODUS: Make lib & include dirs in python modules configurable

e4t:
Shard libraries and include files may not always reside in
CMAKE_INSTALL_PREFIX/lib or CMAKE_INSTALL_PREFIX/include
as these may be configured via ${PROJECT_NAME}_INSTALL_LIB_DIR
or ${PROJECT_NAME}_INSTALL_INCLUDE_DIR.
Use these settings to determine directory names.
Only of these cannot be determined this way, fall back to
'lib' and 'include'

Signed-off-by: Egbert Eich <eich@suse.com>


e4t:

recheck

gsjaardema:
@mvlopri Since you do most of the exodus.py work, are you ok with this suggested change.

gsjaardema:
@e4t  I editted your signing of the CLA and removed the `recheck`.  It looks like the cla-bot is not very smart and expects the comment to be just that one single line...

Thanks for the contribution and thanks for putting up with the extra cruft...

PR #283, opened 2022-02-27
IOSS Assembly: Enable inclusion/exclusion of assemblies and associated element blocks

tokusanya:


tokusanya:
@gdsjaar  I have the same changes on cee-gitlab along with STK unit tests to show that it works

packages/seacas/libraries/ioss/src/exodus/Ioex_BaseDatabaseIO.C
```diff
@@ -95,6 +95,76 @@ namespace {
...
+          if(Ioss::ASSEMBLY == assemblyType) {
+            for (int i = 0; i < assembly.entity_count; i++) {
+              // Find the sub assembly with the same id
+              int64_t subAssemblyId = assembly.entity_list[i];
+              bool found = false;
+              for(size_t j=0; j<m_assemblies.size(); j++) {
+                if(m_assemblies[j].id == subAssemblyId) {
+                  found = true;
+                  update_list_from_assembly_tree(j, list);
+                }
```
gsjaardema, line 137: Do we keep iterating this for loop after finding a matching sub assembly, or should there be a break out of the loop once found is true?

packages/seacas/libraries/ioss/src/exodus/Ioex_BaseDatabaseIO.C
```diff
@@ -95,6 +95,76 @@ namespace {
...
+        }
+      }
+
+      std::sort(assemblyFilterList.begin(), assemblyFilterList.end(), std::less<std::string>());
+      auto endIter = std::unique(assemblyFilterList.begin(), assemblyFilterList.end());
+      assemblyFilterList.resize(endIter - assemblyFilterList.begin());
+    }
+
+  private:
+    AssemblyTreeFilter() = delete;
```
gsjaardema, line 160: Since these are deleted, no benefit to put them down in private section.  Maybe better to put them up by enabled constructor so can see that they are deleted... Keep constructors together...

packages/seacas/libraries/ioss/src/exodus/Ioex_BaseDatabaseIO.C
```diff
@@ -95,6 +95,76 @@ namespace {
...
+    AssemblyTreeFilter(const AssemblyTreeFilter&) = delete;
+
+    Ioss::Region* m_region;
```
gsjaardema, line 163: Initialize to nullptr

packages/seacas/libraries/ioss/src/exodus/Ioex_BaseDatabaseIO.C
```diff
@@ -95,6 +95,76 @@ namespace {
...
+    Ioss::EntityType m_type;
```
gsjaardema, line 164: Initialize to `INVALID_TYPE`

packages/seacas/libraries/ioss/src/exodus/Ioex_DatabaseIO.C
```diff
@@ -1298,6 +1298,10 @@ namespace Ioex {
     m_groupCount[entity_type] = used_blocks;
 
     if (entity_type == EX_ELEM_BLOCK) {
+      if(!assemblyOmissions.empty() || !assemblyInclusions.empty()) {
```
gsjaardema, line 1301: Ioex::ParallelDatabaseIO probably needs similar code in `get_blocks`

gsjaardema (changes requested):
Looks good in general.  A few comments / suggestions.

packages/seacas/libraries/ioss/src/exodus/Ioex_BaseDatabaseIO.C
```diff
@@ -563,6 +633,89 @@ namespace Ioex {
...
+    }
+
+    if(!assemblyInclusions.empty()) {
+      assert(blockOmissions.empty());
+    }
+
+    std::vector<std::string> exclusions;
+    std::vector<std::string> inclusions;
+
+    // Query number of coordinate frames...
```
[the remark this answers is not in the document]
gsjaardema: I do note that the comment is wrong in the original code...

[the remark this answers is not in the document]
tokusanya: Yeah, that was from a copy/paste ... can fix that

packages/seacas/libraries/ioss/src/exodus/Ioex_BaseDatabaseIO.C
tokusanya: No problem. I typically put these constructors in the private section so as to prevent default and copy constructors from being created. Being private also means that you catch potential misuse at compile time rather than at runtime.

packages/seacas/libraries/ioss/src/exodus/Ioex_BaseDatabaseIO.C
tokusanya: I can do that also. As I pointed out earlier, making the object non-default and non-copy constructible means that the only way to create it is through the specified constructor and that does the initialization of the class members. My eclipse IDE flags data members that are not initialized properly

packages/seacas/libraries/ioss/src/exodus/Ioex_BaseDatabaseIO.C
tokusanya: Same as above

packages/seacas/libraries/ioss/src/exodus/Ioex_DatabaseIO.C
tokusanya: Ok, I'll add that too

packages/seacas/libraries/ioss/src/exodus/Ioex_BaseDatabaseIO.C
gsjaardema: Yes, that was the recommended practice prior to `=delete` being added to the language.  The `=delete` will catch potential misuses at compile time.

packages/seacas/libraries/ioss/src/exodus/Ioex_BaseDatabaseIO.C
gsjaardema: Understood.  I like to still initialize them just in case someone changes them to be default or copy constructible or an operator= gets generated or added.

gsjaardema (approved):

PR #285, opened 2022-03-07
TextMesh: block skinning to define sidesets

tokusanya:


gsjaardema (commented):
What is the driver, use-case(s), documentation, and priority for this change?   Looks like lots of added complexity and lots of work expended.  

Too much code to review since hard to see what changes are for the functionality added and which are just moving things around.

Also, I don't see any documentation for what the feature is and how it works. Either in internal commented code or external user-facing documentation...

PR #286, opened 2022-03-07
TextMesh: Block skinning for sideset definition

tokusanya:
@gsjaardema Not sure why this isn't an automatic merge. Iotm_TextMeshUtil.h contains all the previous assembly changes and the new skinning code

### DIFF
--- a/packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMesh.h
+++ b/packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMesh.h
@@ -25,9 +25,9 @@
 
 namespace Iotm {
   using Topology       = TopologyMapEntry;
-  using TextMeshData   = text_mesh::TextMeshData<int64_t, TopologyMapEntry>;
-  using ElementData    = text_mesh::ElementData<int64_t, TopologyMapEntry>;
-  using SidesetData    = text_mesh::SidesetData<int64_t, TopologyMapEntry>;
+  using TextMeshData   = text_mesh::TextMeshData<int64_t, Topology>;
+  using ElementData    = text_mesh::ElementData<int64_t, Topology>;
+  using SidesetData    = text_mesh::SidesetData<int64_t, Topology>;
   using NodesetData    = text_mesh::NodesetData<int64_t>;
   using AssemblyData   = text_mesh::AssemblyData<int64_t>;
   using Coordinates    = text_mesh::Coordinates<int64_t>;
@@ -36,11 +36,6 @@ namespace Iotm {
   using SideBlockInfo  = text_mesh::SideBlockInfo;
   using SplitType      = text_mesh::SplitType;
   using AssemblyType   = text_mesh::AssemblyType;
-
-  inline std::ostream &operator<<(std::ostream &out, const TopologyMapEntry &t)
-  {
-    return out << t.name();
-  }
 
   struct BlockPartition
   {
--- a/packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshTopologyMapping.h
+++ b/packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshTopologyMapping.h
@@ -17,47 +17,52 @@
 #include <utility> // for pair
 #include <vector>  // for vector
 
+#include <assert.h>
+
 #include "Ioss_ElementTopology.h"
 #include "Ioss_StandardElementTypes.h"
 
 #include "Iotm_TextMeshUtils.h"
 
 namespace Iotm {
-  struct TopologyMapEntry
+  class TopologyMapEntry
   {
+  public:
+    using Ordinal = uint16_t;
+    using Permutation = uint8_t;
+
+    static constexpr Ordinal InvalidOrdinal = 65535;
+    static constexpr Permutation InvalidPermutation = 128;
+
     using DimensionArray = bool[4];
 
-    unsigned int           id;
-    Ioss::ElementTopology *topology;
-
-    // Defines what spatial dimension the topology is valid on
-    DimensionArray validSpatialDimensions;
-
     TopologyMapEntry()
-        : id(Ioss::ElementTopology::get_unique_id(Ioss::Unknown::name)),
-          topology(Ioss::ElementTopology::factory(Ioss::Unknown::name))
+        : id(Ioss::ElementTopology::get_unique_id(std::string(Ioss::Unknown::name))),
+          topology(Ioss::ElementTopology::factory(std::string(Ioss::Unknown::name))),
+          isShell(false),
+          numPermutations(0),
+          numPositivePermutations(0),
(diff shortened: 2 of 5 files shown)


gsjaardema:
What is the driver, use-case(s), documentation, and priority for this change? Looks like lots of added complexity and lots of work expended.   For example, why does text mesh need a block skinning mode?

Too much code to review since hard to see what changes are for the functionality added and which are just moving things around.

Also, I don't see any documentation for what the feature is and how it works. Either in internal commented code or external user-facing documentation...

tokusanya:
This was a feature that I had worked as a personal project a few months ago but never got around to pushing. It was intended as another way to define sidesets through block skinning and is based on a simplified version of the skinning code in StkMesh which constructs a face-face adjacency graph for determining neighbors. Those element sides without a neighbor are candidates for entry into a sideset. In that sense, the title of the PR is probably a misnomer since all it's doing is really figuring out the mesh adjacency.  It properly handles all the one-off cases for shell connections as well as stacked elements which have come up a lot in many STK use cases. I currently have the STK version of it up on the cee-gitlab. I wouldn't say that there is any sort of priority on it since I wrote this on my own time with the  intention to make sideset definitions easier.

The documentation for the feature is as before for the options in the print_help_message() function which has been modified to show the usage. The face-face connectivity graph is not meant to be user-facing ... it is internal implementation based on the STK elem-elem graph that we use internally for all of our connectivity needs. It is a lot of code to take in at once to enable one user-facing feature but I wonder if splitting this off into a separate impl file might help. 

The STK team tries as much as possible to follow Uncle Bob coding standards by not adding comments and making sure that functions and variables are appropriately named to enable better understanding of the code. This is why we have lots of unit tests which essentially serve as documentation. I could add lots of comments to the changes so it's easier to follow but I would have to be more careful in maintaining the STK version of it (if you remember, there is a STK version of TextMesh also that uses the same header file). I would actually like to discuss what the coding standards in SEACAS are at some point ..... 

Please let me know your thoughts...I can always put off pushing this until a time when this might be more useful for users. It's something I see coming in handy when the exo->textmesh story for Mark Merewether comes around.

gsjaardema:
Not sure I'm asking for "lots of comments" but I think there are inherent assumptions that should be commented.  As a simple example, there is the assumption that all positive permutations are listed first, but that assumption isn't stated anywhere and I don't think that it is checked.  If I add a new element how do I know about that assumption and possibly others.    As another example, are there any assumptions on the ordering of the positive and negative permutations.  Is "012, 120, 201" as valid as "012 201 120" or "201 012 210" for the Tri3 elements...

At a minimum, it seems that splitting out the assembly-related changes and from the sideset changes would be worthwhile.  I'm still not sold on the addition of all the topology-related code inside TextMesh...

tokusanya:
As you surmised, there are several inherent assumptions that come from the STK side of things. The implementation of the face-face adjacency graph (in STK and thus here) is is heavily dependent on node matching on element sides and this is made possible by the use of node permutations on elements and faces. I looked through Ioss_ElementTopology but there was only minimal support for this which is why I had to extend the Iotm_ElementTopologyMapping class to handle all the use cases that are supported in stk_topology.

To address the specific question you had regarding the ordering of the positive and negative permutations (within their groups), there isn't ... the only thing required is that if there are N positive permutations and M total permutations (M >= N), then the first N permutations of the total permutations are the positive ones and the remaining M-N are all negative permutations. stk_topology always has the very first positive permutation correspond to the default Exodus node ordering.

gsjaardema:
I understand uncle bob's recommendations, but I don't think that he is suggesting that there be no comments and a person has to look into the implementation of a function to determine what it actually does and what assumptions or requirements it has.  An example is the listing of positive permutations prior to negative permutations -- that is an assumption throughout the code, but it is never stated and can only be determined by looking at the implementation of the code.

There are some other situations similar -- 
* why would a `tet8` element only have 1 permutation, but a `tet4` and a `tet10` both have 12 permutations?  There is nothing in the function naming or anything else I can see that tells me why. 
* why do some types have the `set_side_topologies` listed and some don't.
* Why do all higher-order types need to have their permutations specified; it seems like just using the corner vertices of the lowest order element would still give you the information you need for the skinning, and would be much simpler coding and adding of new elements...  For example, a single listing of hex8 permutations would support skinning of all higher-order hex elements (9, 16, 20, 27, 32, 64)
* Why is a `line_3` and a `line_3_1d` needed instead of just a `line_3` that is valid for 1, 2, and 3D? 
* What does the function name `strip` tell me about what it does?
* The function `is_number` only really supports non-negative integers, but I need to read the implementation to determine that.
* `assign_id_from_standard_name` and `assign_id_for_non_standard_name` -- what is a standard vs non-standard name (I konw from reading the implementation, but that usage doesn't match anything else in IOSS and a simple comment would keep me from having to read the source)

[Note that I am not necessarily asking for explicit answers to the above questions; I'm just pointing out that there are many assumptions and requirements in the code that can not be determined just by reading the code or the function names and they could be easily explained by a comment explaining why the code is written the way it is]

I have read Uncle Bob for a couple decades and follow some of his recommendations and also agree that a function name should give an idea of what the function does, but I don't think that eliminating all comments is the way to go.  If noone is going to use your code except for you, then maybe that is OK...

I am fine with adding additional functionality to IOSS if it is needed.  I'm not sure that text mesh should have all of the element topology information embedded in it since it is not useable by any other class if it is there.  Maybe that functionality should be separated out and maybe even extend the Ioss_ElementTopology classes (they already have the side topology information).

tokusanya:
I get where you are coming from ... the STK team decided to literally take Uncle Bob to heart so we kinda went a little bit on the extreme side of things ... particularly the "any comment is a failure" mantra. One of the reasons our code reviews go so long is because we nitpick details to death so that comments can be eliminated. It's exhausting sometimes but it does help ... some of the feedback you've given me so far has already made me rethink the naming of my variables and functions. I've been waiting for a code review from the STK team on this but they've been too busy with other things so your comments are really helpful. It's one of the reasons I wanted to discuss SEACAS coding standards so that I have a better idea of how to modify the code.

As for the topology based functionality, these are all things that already exist in stk_topology and the permutation data is copy/paste from STK. These were all worked out when they were first created by the original STK team way back when. I actually wondered when I was writing the code if Ioss_ElementTopology would benefit by extending it to include permutations but I wasn't sure if that was a closed feature set or if the rest of IOSS would have benefited so I kept it all within TextMesh. If you think that's a worthwhile endeavour, I can look into that.

gsjaardema:
> if Ioss_ElementTopology would benefit by extending it to include permutations but I wasn't sure if that was a closed feature set or if the rest of IOSS would have benefited so I kept it all within TextMesh. If you think that's a worthwhile endeavour, I can look into that.

I would be fine with that.  Also it might be worthwhile to see if the skinning code could be standalone utility for use by all meshes and an alternative to the existing skinning capability in ioss...

PR #289, opened 2022-03-14
Assembly: Fix logic error when omitting members

tokusanya:
Also added a small format change

### DIFF
--- a/packages/seacas/libraries/ioss/src/exodus/Ioex_BaseDatabaseIO.C
+++ b/packages/seacas/libraries/ioss/src/exodus/Ioex_BaseDatabaseIO.C
@@ -144,8 +144,8 @@ namespace {
               if(!found) {
                 std::ostringstream errmsg;
                 fmt::print(errmsg,
-                    "ERROR: Could not find sub-assembly with id: {} and name: {}"
-                    "       [{}]\n", assembly.id, assembly.name);
+                    "ERROR: Could not find sub-assembly with id: {} and name: {}",
+                    assembly.id, assembly.name);
                 IOSS_ERROR(errmsg);
               }
             }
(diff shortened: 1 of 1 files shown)


gsjaardema:
I don't see the logic error change; only two format changes...  Is something missing in the commit?

tokusanya:
Line 775 had a logic error where the GroupingEntity is not a nullptr but omitted ...this would cause it to throw which was not correct. It should only throw if the GroupingEntity is null

packages/seacas/libraries/ioss/src/exodus/Ioex_BaseDatabaseIO.C
```diff
@@ -771,10 +771,11 @@ namespace Ioex {
             assem->add(ge);
             num_added_entities++;
           }
-          else {
+
+          if (ge == nullptr) {
```
gsjaardema, line 775: I think the overall logic here would be simpler if this `ge == nullptr` test were put right after the `auto *ge = ...` line.  Right now we have a test: 
```
 if (ge != nullptr && !Ioss::Utils::block_is_omitted(ge)) {
```
and then a test: 
```
 if (ge == nullptr) {
```
If the second test was first, then the first test would just be:
```
if (!Ioss::Utils::block_is_omitted(ge))
```

gsjaardema (commented):
Is ok to merge as is, but I think the logic is easier to read if the two tests are reordered.

packages/seacas/libraries/ioss/src/exodus/Ioex_BaseDatabaseIO.C
tokusanya: That sounds good

PR #291, opened 2022-03-17
TextMesh: Face adjacency graph to enable block skinning for sideset definition

tokusanya:
Refactored code by splitting into files for each TextMesh capability


tokusanya:
@gsjaardema I think you might have merged the old PR that I had closed ...this is the latest one with all new changes

gsjaardema:
@tokusanya I merged #292 which looks like it was just opened 5 days ago.  Looked open when I merged it... #285 and #286 were closed and not merged as best I can tell...

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshFuncs.h
```diff
@@ -0,0 +1,140 @@
...
+{
+  std::transform(str.begin(), str.end(), str.begin(), ::toupper);
+}
+
+inline void convert_to_lower_case(std::string &str)
+{
+  std::transform(str.begin(), str.end(), str.begin(), ::tolower);
+}
+
+inline bool is_positive_number(const std::string &str)
```
gsjaardema, line 98: pedantic: is non-negative, but ok with positive...

PR #296, opened 2022-03-31
APREPRO: Add detail to the --debug help message

mvlopri:
This will make it more clear that passing --debug will keep around
the temporary files used for arrays by aprepro.


gsjaardema:
Thanks.

PR #306, opened 2022-05-24
EXODUS.PY: Add the get_block_id_map method

mvlopri:
This adds functionality to add being able to easily find the
elements contained within EX_ELEM_BLOCK or other type without
keeping track of global and internal indicies.


gsjaardema:
Thanks, looks good to me.

PR #310, opened 2022-06-10
EXODUS.PY: Add context manager functionality to the exodus class

mvlopri:
This will allow the use of with statements for opening and closing
exodus files, including methods which return exodus class objects,
such as exodus.copy.


mvlopri:
This need an error handling patch before it merges.

PR #313, opened 2022-06-14
EXODUS.PY: Update documentation to reflect contextmanager changes

mvlopri:
Exodus objects can now be used through with statements.
This change makes the documented usages reflect this change
so it can be easily found.


mvlopri:
Do you think this makes it clearer for users? If so, I will make a PR for the docs repository as well.

gsjaardema:
Yes, I think this makes it clearer and I think it is more pythonic...

PR #315, opened 2022-06-15
Avoid Real[N] in registry if ignoring such fields

### LINKED ISSUE #314: Unable to ignore real[N] fields after not ignoring once.
Odd behavior when the `IGNORE_REALN_FIELDS` property is toggled on/off and then turned on again.

-> ignore_realn_fields = on
   names field_1, field_2, field_3 will result in three scalar fields.

-> ignore_realn_fields = off
   names field_1, field_2, field_3 result in 3-component array

-> ignore_realn_fields = on
   names field_1, field_2, field_3 will still result in 3-component array!

No matter what, it behaves as if that property is off. After every toggle, I create a new `DatabaseIO` object with updated properties and start from scratch.

jspanchu:
The state of `ignore_realn_fields` is not considered when handing out a variable from the factory registry. Fix #314

### DIFF
--- a/packages/seacas/libraries/ioss/src/Ioss_VariableType.C
+++ b/packages/seacas/libraries/ioss/src/Ioss_VariableType.C
@@ -186,6 +186,10 @@ namespace Ioss {
     bool match = false;
     for (const auto &vtype : registry()) {
       auto *tst_ivt = vtype.second;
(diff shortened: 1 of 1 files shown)


jspanchu:

jspanchu:
recheck

jspanchu:
recheck

PR #324, opened 2022-07-29
Add test for put_qa_records to help serve as documentation of use

mvlopri:


packages/seacas/scripts/tests/test_exodus3.py
```diff
@@ -129,6 +129,13 @@ def test_get_assemblies(self):
         self.maxDiff = None
         self.assertEqual(str(expected), str(assemblies))
 
+    def test_put_qa(self):
+        new = [("name","yeet","woot","scoot")]
```
gsjaardema, line 133: Just to provide a little more realism.  The QA records are supposed to be:

- code name
- code qa
- execution date,
- execution time.

Maybe do a "name", "1.0.0-1", "20220801", "12:34:56"

Doesn't make much difference, but at least follows the form of what a qa record is and if there were ever any validation of the qa records, then those values would pass the validation...

gsjaardema: Had some time while waiting on a compile, so just did the change myself.

PR #334, opened 2022-10-18
SEACAS: Remove listing of undefined TPL deps (trilinos/Trilinos#11152)

bartlettroscoe:
Matches changes in Trilinos PR https://github.com/trilinos/Trilinos/pull/11154 which fixes trilinos/Trilinos#11152.


packages/seacas/libraries/ioss/cmake/Dependencies.cmake
```diff
@@ -6,7 +6,7 @@ TRIBITS_PACKAGE_DEFINE_DEPENDENCIES(
 else()
 TRIBITS_PACKAGE_DEFINE_DEPENDENCIES(
   LIB_OPTIONAL_PACKAGES SEACASExodus Pamgen Zoltan Kokkos
-  LIB_OPTIONAL_TPLS HDF5 CGNS ParMETIS Faodel Cereal DLlib Pthread DataWarp ADIOS2 Catalyst2 GTest
+  LIB_OPTIONAL_TPLS HDF5 CGNS ParMETIS DLlib Pthread ADIOS2
```
gsjaardema, line 9: I'm not sure about this one.  If I enable Faodel, don't I need to have this listed here?  Same with Catalyst2.  There are also some tests that use GoogleTest if it is available and Datawarp on some Cray systems.  

Probably a correct change; just trying to understand the details.

bartlettroscoe: @gsjaardema, the problem is that these TPLs are not listed in the file [Trilinos/TPLsList.cmake](https://github.com/trilinos/Trilinos/blob/develop/TPLsList.cmake) so when SEACAS is configured under Trilinos, these TPLs are not defined.  I can see that they are listed in the file [seacas/TPLsList.cmake](https://github.com/sandialabs/seacas/blob/master/TPLsList.cmake) so they are defined.

There is one of two options here.

One option is it make the listing of these TPL dependencies in these Dependencies.cmake files dependent on if they are running from the SEACAS project or not.  But that would add a good bit of conditional logic to these files (but one could set up vars that do this in a more smooth way).

A second option is to copy the lines:

https://github.com/sandialabs/seacas/blob/d04c860e99ffe911d26b7b286f7fefee993f69d1/TPLsList.cmake#L64

https://github.com/sandialabs/seacas/blob/d04c860e99ffe911d26b7b286f7fefee993f69d1/TPLsList.cmake#L74-L75

https://github.com/sandialabs/seacas/blob/d04c860e99ffe911d26b7b286f7fefee993f69d1/TPLsList.cmake#L77

into the [Trilinos/TPLsList.cmake](https://github.com/trilinos/Trilinos/blob/develop/TPLsList.cmake) file so these TPLs are defined when Trilinos gets configured.

But then there is the problem with the GTest TPL.  That TPL is called 'gtest' in Trilinos:

https://github.com/trilinos/Trilinos/blob/3428aa23090a36102aeb3767f774c5879600f0ec/TPLsList.cmake#L127

```
  gtest           "cmake/TPLs/"    EX
```

Because there is already a "GTest" TriBITS internal package defined for Trilinos:

https://github.com/trilinos/Trilinos/blob/3428aa23090a36102aeb3767f774c5879600f0ec/PackagesList.cmake#L63

```
  Gtest                 commonTools/gtest                 PT
```

we can't just copy that over.

Considering all of this, this is what I would suggest:

1. Copy the lines from [seacas/TPLsList.cmake](https://github.com/sandialabs/seacas/blob/d04c860e99ffe911d26b7b286f7fefee993f69d1/TPLsList.cmake) for these 4 TPLs to  [Trilinos/TPLsList.cmake](https://github.com/trilinos/Trilinos/blob/develop/TPLsList.cmake) (but not for GTest).  That way, customers of Trilinos could be enabling these TPLs as well when building SEACAS under Trilinos.  (That seems like the right thing to do.)

2. For the SEACAS GTest TPL, define a var `set(SEACAS_GTest_TPL_name "GTest")` in the file [seacas/ProjectName.cmake](https://github.com/sandialabs/seacas/blob/master/ProjectName.cmake) and then change the listing of 	`GTest` in these SEACAS dependency files to be `${SEACAS_GTest_TPL_name}`.  So when building under the SEACAS TriBITS project, the `GTest` TPL would be defined but when SEACAS under the Trilinos TriBITS project, it would not be defined.

Note that before adding this check, when building SEACAS under Trilinos, TriBITS was just silently ignoring these undefined TPLs and not adding them to the list of dependencies internally.  (It is just they ever even existed.  This was a "robust" implementation but does not catch mistakes like this.)

gsjaardema: Thanks.  I will make the suggested changes.

PR #335, opened 2022-10-19
Export ioss symbols

mathstuf:
---
I want to split off some of these things, but want CI to chew on the whole thing at once first.


gsjaardema:
Looks like the new export include files weren't included.

mathstuf:
Yeah, I'll need to debug that; it works locally (for exodus and cgns at least…) and the other deps aren't building here for whatever reason.

mathstuf:
There's a backwards edge in the build graph hidden in `Ioss_Decomposition.C`

mathstuf:
`Ioss_Transform.h` defines APIs in the `Iotr` namespace and implemented in `transform/` (the `Iotr` library). This doesn't work. How best to fix this design?

packages/seacas/libraries/ioss/src/Ioss_Decomposition.C
```diff
@@ -20,10 +20,6 @@
 extern "C" int Zoltan_get_global_id_type(char **name);
 #endif
 
-#if defined(SEACAS_HAVE_CGNS)
```
gsjaardema, line 23: Not sure how that ended up there...  Does not look like it is needed for anything.

mathstuf:
I reorganized your diffs into more sensible commits and fixed(?) the next MSYS error.

PR #337, opened 2022-10-20
Warning fixes

mathstuf:


mathstuf:
I can break this up into smaller PRs if you prefer.

packages/seacas/libraries/ioss/src/Ioss_Beam2.C
```diff
@@ -79,13 +79,15 @@ int Ioss::Beam2::number_nodes_face(int face) const
 {
   // face is 1-based.  0 passed in for all faces.
   assert(face >= 0 && face <= number_faces());
+  (void)face;
```
gsjaardema, line 82: Can we use a macro or something here to make it more visible what is happening -- UNUSED(face) or something like that.  Then can tailor for different compilers as I think there ar some that still warn even with `(void)face;`.  Can use C++ attributes `[[maybe_unused]]` (?) in the macro...  

I know macros are frowned upon, but makes it easier to tailor for multiple compilers...

packages/seacas/libraries/ioss/src/Ioss_Utils.C
```diff
@@ -580,15 +580,15 @@ namespace {
         // It is possible that the first name(s) that match with two
         // suffices have a basename that match other names with only a
         // single suffix lc_cam_x, lc_cam_y, lc_sfarea.
-        for (int i = index + 1; i < num_names; i++) {
-          char                    *tst_name = names[i];
+        for (int id = index + 1; id < num_names; id++) {
```
gsjaardema, line 583: Good catch.  I thought I had fixed this, but definitely didn't..  Maybe I accidentally fixed on a branch...

packages/seacas/libraries/ioss/src/Ioss_Beam2.C
mathstuf: That's a C++17 attribute (and belongs on the parameter itself IIRC). Where should this macro live?

gsjaardema: That works.  Or `Ioss_CodeTypes.h`

gsjaardema: Is Kitware using C++17 yet?  Sierra has moved to C++17 and Trilinos will soon, so C++17 will soon start appearing in the source code.  I can delay if that will adversely affect your builds though.

mathstuf:

mathstuf:
recheck

packages/seacas/libraries/ioss/src/Ioss_Beam2.C
mathstuf: VTK is still firmly in C++11. I think VTK-m will push to C++14 in the short term. Our vendoring has a commit that makes things C++11-compatible though; it's not too much  a deal as long as that match doesn't get too crazy. https://gitlab.kitware.com/third-party/seacas/-/commit/42fa9767d657ff14ca2b9134cfa3d335aea5323f

packages/seacas/libraries/ioss/src/cgns/Iocgns_DatabaseIO.C
```diff
@@ -1819,9 +1818,8 @@ namespace Iocgns {
 
     // Create a lambda to eliminate some duplicate code in coordinate outputs...
     auto coord_lambda = [&data, &first,
-                         base](const char *ordinate, int cgns_file_ptr,
-                               const std::vector<CGNSIntVector> &block_local_node_map,
-                               int                               myProcessor) {
+                         base, this](const char *ordinate, int cgns_file_ptr,
+                               const std::vector<CGNSIntVector> &block_local_node_map) {
```
gsjaardema, line 1822: should `block_local_node_map` be changed to `m_blockLocalNodeMap` and accessed through `this` now also?  Also potentially `get_file_pointer()`.  

Probably something to look at later instead of on this PR.

packages/seacas/libraries/ioss/src/cgns/Iocgns_DatabaseIO.C
```diff
@@ -1888,14 +1885,14 @@ namespace Iocgns {
           // End of lambda...
           // ========================================================================
 
-          blk_coord_lambda("CoordinateX", 0, get_file_pointer(), myProcessor);
+          blk_coord_lambda("CoordinateX", 0, get_file_pointer());
```
gsjaardema, line 1888: Could potentially access `get_file_pointer()` internal to lambda now that `this` is being passed in...  (something to look at for later)

gsjaardema (approved):
Looks good.  Thanks for doing this; lots of work.

PR #338, opened 2022-10-20
Namespacing

mathstuf:


mathstuf:
Why a Python linter would care about C++ only changes…no idea.

gsjaardema:
> Why a Python linter would care about C++ only changes…no idea.

I should have done these changes on a branch.  Added a python linter github action and it has issues with some of the current code.  Sorry for the confusion.

PR #340, opened 2022-10-27
EXODUS.PY: Add a test for the copy_file method

mvlopri:
The interface for copy_file was exposed as part of merge request #325. This adds a test for it to the unittests.


gsjaardema:
Thanks.  Looks good.

PR #345, opened 2022-11-09
EXOMERGE_UNIT_TEST: Fix linting issue in exomerge unit tests

mvlopri:


gsjaardema:
Looks like there is a problem with one of the tests -- it fails an assertion and that assertion looks like it is a valid test. The test involves merging nodes, so if any nodes get merged, then the final count should be less than the original count.

I can take a closer look if the fix isnt' clear.

PR #350, opened 2022-11-18
EXODUS.PY: Refactor various code patterns

mvlopri:
These changes are intended to make exodus.py a bit more pythonic with a few common patterns.
Using newer sytax for formating strings
Using list comprehensions in place of list building for loops Not creating variables just to return them
Preventing latent bugs in mutable default arguments Prenting memory leaks through use of with instead of open Various other refactors


gsjaardema:
Looks good.  There seems to be one flake8 error.

PR #363, opened 2023-01-18
Catalyst vector outputs

tjotaha:
Changes Catalyst plugin to output vtkPartitionedDataSetCollection from vtkMultiBlockDataSet. This change will provide better compatibility with the ParaView IOSS reader output in the Paraview GUI and improve combability of exported Catalyst scripts.

Variables with multiple components are grouped into a single vector rather than split into multiple single component vectors when output to Catalyst.

Adds version checking between SEACAS IOSS and the Catalyst plugin when the Catalyst plugin is loaded.

Updated version of Phactori Python and associated tests.


gsjaardema:
LGTM

PR #364, opened 2023-02-01
Catalyst Exodus Database

tjotaha:
Provide implementations for all put_field_internal methods for unsupported IOSS entity types. These
implementations return num_to_get.

Fix return types in ioss2catalyst test code.

### DIFF
--- a/packages/seacas/libraries/ioss/src/visualization/exodus/Iovs_exodus_DatabaseIO.C
+++ b/packages/seacas/libraries/ioss/src/visualization/exodus/Iovs_exodus_DatabaseIO.C
@@ -38,8 +38,8 @@ namespace Iovs_exodus {
       : Ioss::DatabaseIO(region,
                          Iovs::Utils::getInstance().getDatabaseOutputFilePath(filename, props),
                          db_usage, communicator, props),
-        isInput(false), singleProcOnly(false), doLogging(false),
-        nodeBlockCount(0), elementBlockCount(0)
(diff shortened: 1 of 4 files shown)


packages/seacas/libraries/ioss/src/visualization/exodus/Iovs_exodus_DatabaseIO.C
```diff
@@ -567,6 +565,73 @@ namespace Iovs_exodus {
...
+                                         void * /*data*/, size_t data_size) const
+  {
+    return get_num_to_get(field, data_size);
+  }
+  int64_t DatabaseIO::put_field_internal(const Ioss::Blob * /*bl*/, const Ioss::Field &field,
+                                         void * /*data*/, size_t data_size) const
+  {
+    return get_num_to_get(field, data_size);
+  }
+  size_t DatabaseIO::get_num_to_get(const Ioss::Field &field, size_t data_size) const
```
gsjaardema, line 630: Could probably put this as a `namespace{}` function in the current file.  Does not need to be  a class member function.

gsjaardema: But not needed for this PR.   Could modify later if wanted.

tjotaha: Okay. Will push a change for this.

PR #368, opened 2023-02-14
EXODUS.PY: Fix element attribute query by name

gsjaardema:
Fix the `__ex_get_one_attr` function to correctly query the number of entities (nodes, elements) in the set or block.  Still not as generic as it should be and we should probably refactor into a separate function the code returning `numVals` as it is used in at least two places in the code.  

@mvlopri

### DIFF
--- a/packages/seacas/scripts/exodus3.in.py
+++ b/packages/seacas/scripts/exodus3.in.py
@@ -2216,12 +2216,23 @@ class exodus:
         """
 
         numVals = 0
-        if objType == 'EX_NODAL':
+        if objType == "EX_NODAL":
             numVals = self.num_nodes()
-        elif objType == 'EX_ELEM_BLOCK' or objType == 'EX_FACE_BLOCK' or objType == 'EX_EDGE_BLOCK':
-            (_elemType, numElem, _nodesPerElem, _numAttr) = self.__ex_get_block(objType, entityId)
+        elif (
+            objType == "EX_ELEM_BLOCK"
+            or objType == "EX_FACE_BLOCK"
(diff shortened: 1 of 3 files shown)


packages/seacas/scripts/exodus3.in.py
```diff
@@ -5490,17 +5490,30 @@ def __ex_put_one_attr(self, objType, elemBlkID, attrIndx, Attr):
...
+    def __ex_get_one_attr(self, objType, entityId, attrIndx):
+        entity_id = ctypes.c_longlong(entityId)
         obj_type = ctypes.c_int(get_entity_type(objType))
         attr_index = ctypes.c_longlong(attrIndx)
-        inqType = ex_inquiry_map(ex_obj_to_inq(objType))
-        num_objs = ctypes.c_int(self.__ex_inquire_int(inqType)).value
-        attrib = (ctypes.c_double * num_objs)()
+
+        numVals = 0
+        if objType == 'EX_NODAL':
```
mvlopri, line 5499: I think it would be a good idea to separate this if-elif chain out as a method for a (mostly) generic way of getting the number of elements. Long term I think it would be best to invert these methods to be part of classes for EX_*.  I think this a good change and I will try to write a test for it soon as well.

gsjaardema: I split that out into a `get_entity_count(self, objType, entityId)` function and replaced the 3 uses of it in the code with a call to the function.  Also added some additional types to the function. 

Made a more generic version of the attribute value function which should work with all sets/blocks.  Tested to make sure it still worked on the specific case that triggered this change, but definitely needs more testing.

PR #372, opened 2023-03-15
Ioss zero copy

gsjaardema:
Prototype implementation to support the zero-copy API discussed in #282 

@tjotaha @mathstuf @utkarshayachit

### DIFF
--- a/packages/seacas/applications/slice/Slice.C
+++ b/packages/seacas/applications/slice/Slice.C
@@ -139,8 +139,8 @@ namespace {
     // The chain / line data will be stored as an element map...
     const auto &blocks = region.get_element_blocks();
     for (const auto &block : blocks) {
-      Ioss::Field field{"chain", region.field_int_type(), "Real[2]", Ioss::Field::MAP};
-      field.set_index(1);
+      auto field =
+          Ioss::Field("chain", region.field_int_type(), "Real[2]", Ioss::Field::MAP).set_index(1);
       block->field_add(field);
     }
   }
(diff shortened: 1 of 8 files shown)


mathstuf:
@spyridon97

spyridon97:
Ι am not an IOSS expert, but the changes seem good to me.

The only thing that i would change is to convert the return type of get_field_data to an enum that inherits int64_t (to maintain backwards - compatibility) with specific enum values that are easily interpretable.

utkarshayachit:
+1

gsjaardema:
> An IOSS database that supports zero_copy reads needs to provide an implementation of this new method in Ioss::GroupingEntity?
> 
> ```c++
> int64_t Ioss::GroupingEntity::internal_get_field_data(const Ioss::Field &field, void ** /*data*/,
>                                                       size_t * /*data_size*/) const
> ```
Yes.

> 
> Then the code that reads from this IOSS database needs to specify all of the Ioss::Field() objects being added to an Ioss::GroupingEntity as being zero_copy (code below from ParaView's IOSS reader)?
> 
> ```c++
>         block->field_add(
>           Ioss::Field(std::get<0>(field), std::get<1>(field), "scalar", role, elementCount));
> ```
Yes, the DatabaseIO object that is going to support zero-copy reads would specify which fields that it is adding would be zero_copy_enabled.  A client of that database could access the field data either way -- provide a memory space for the data to be stored in, or ask for the pointer to the memory space managed by the DatabaseIO object.
> 
> Should the Ioss::Field constructor contain a parameter to set zero_copy to true?
Possibly.  I had thought about that but didn't decide either way.  I will take a look at the Ioss::Field constructor(s) and see how cleanly that could be done.  An optional parameter would make it easy, or another constructor...

gsjaardema:
@tjotaha 

> Should the Ioss::Field constructor contain a parameter to set zero_copy to true?

How about something like this:
```
        block->field_add(
          Ioss::Field(std::get<0>(field), std::get<1>(field), "scalar", role, elementCount).set_zero_copyable());
```
Looking at the current constructors, there are already some optional parameters, so adding yet another possibility of being able to pass a TRUE/FALSE parameter or enum value for setting zero copy status might get really confusing.  

With the above idiom, it should give almost the same brevity, but not the confusion or possible parse issues...

tjotaha:
Do we need a Boolean method on IODatabase to indicate that the database supports zero-copy?

PR #373, opened 2023-03-30
Allow modern Netcdf without HDF5

ibaned:
NetCDF can be built without HDF5,
and this change allows that to work
with the "modern" options


ibaned:
This is superceded by other changes

PR #374, opened 2023-03-31
remove modern HDF5 version check

ibaned:
HDF5 exports a version compatibility
file that requires that the major
*and minor* version numbers agree
to be compatible.
Thus, if we give version 1.10 to
find_package, it will not accept
HDF5 1.13 for example.
Since the "modern" mode is opt-in,
it may be easiest just to remove
this minimum version since it was
just a guess anyways.

@bartlettroscoe


ibaned:

gsjaardema:
Failing test has nothing to do with this commit...

PR #381, opened 2023-05-11
Bugfix for CMakeLists.txt

mvlopri:
This allows the use of the full SEACAS package to be included as a cmake_add_subdirectory from a CMake project above it.


mvlopri:
@sebrowne  @bartlettroscoe

bartlettroscoe:
The base CMakeLists.txt file does not get included in the TriBITS build of the SEACAS package.  Only the packages/seacas/CMakeLists.txt file gets included in a Trilinos build of SEACAS.

PR #388, opened 2023-05-24
Fix zc field api

gsjaardema:
Refactor the zero-copy field api so it can actually be used.  Partial implementation here.


gsjaardema:
@tjotaha This is partially what is needed.  Still need some work on changing a field to be zc-enabled and haven't got any tests of actually providing a ZC-enabled databaseIO...

gsjaardema:
Still need to generate a test Ioss databaseio that supports zero-copy and can test this, but comitting so users can see if it works on their code.

PR #389, opened 2023-05-25
IOSS: Allow disabling ioss applications (io_shell, ...)

gsjaardema:
Setting the symbol `${CMAKE_PROJECT_NAME}_ENABLE_SEACASIoss_Applications` to OFF/FALSE will disable the building of io_shell, io_info, io_modify, cgns_decomp, ...  All of the applications in libraries/ioss/src/main


glhenni:
Perfect. Thanks for cranking this out so fast Greg!

PR #394, opened 2023-07-05
Add ex_get_var_time and tests

mvlopri:


packages/seacas/scripts/exodus3.in.py
```diff
@@ -2313,6 +2313,43 @@ def get_entity_count(self, objType, entityId):
...
+        block, variable name, and range of time steps
+
+        >>> evar_vals = exo.get_variable_values_time('EX_ELEM_BLOCK', elem_blk_id,
+        ...                                            evar_name, start_step, end_step)
+
+        Parameters
+        ----------
+        objType   : ex_entity_type
+            type of object being queried
+        entityId : int
```
gsjaardema, line 2328: I think that this `entityId` is the index of the particular node or element within all of the nodes/elements, not a specific node or element block id.  For example, If the database has 10 timesteps and you want the values of the PRESSURE on the 10th element, you would do something like:
```
    evar_10_vals = exo.get_variable_values_time('EX_ELEM_BLOCK`, 10, 'PRESSURE', 1, 10)
```
This is more of an entity-based API function instead of a block/set-based API function ...

mvlopri: The part I found odd is that entityId used for ex_get_var returns the expected value for the variable. Is the entityId of ex_get_var different than the entityId of ex_get_var_time?

gsjaardema: Yes, `ex_get_var` returns the values for all elements/entities in the specified element block or nodeset or sideset at a specific time.  `ex_get_var_time` returns the values for multiple times for a single node or element.  

Sort of a "thread through time" set of values for one node/element instead of a "slab of values for all nodes/elements in a block/set at a specific time.

mvlopri: So, ex_get_var should say blockId instead? That makes more sense given the documentation comments but I think we should change that variable name then.

They my question would be, how do I get the entityId from blockId or is there another way to get that for the "EX_ELEM_BLOCK" objects?

gsjaardema: The comment on line 2329 is incorrect.  Should be *INDEX* and not *ID*

packages/seacas/scripts/exodus3.in.py
```diff
@@ -5781,6 +5818,17 @@ def __ex_get_var(self, timeStep, varType, varId, blkId, numValues):
             var_vals)
         return var_vals
 
+    def __ex_get_var_time(self, varType, varId, entityID, start_step, end_step):
+        s_step = ctypes.c_int(start_step)
+        e_step = ctypes.c_int(end_step)
+        var_type = ctypes.c_int(get_entity_type(varType))
+        var_id = ctypes.c_int(varId)
+        entity_id = ctypes.c_longlong(entityID)
+        time_range = range(start_step, end_step + 1)
```
gsjaardema, line 5827: The use of `time_range` and then using the length of it seems more convulted than just `num_steps = end_step - start_step + 1` and then `var_vals = (ctypes.c_double * num_steps)()`

mvlopri: I agree, that does seem like the simpler option.
Change has been made in latest push.

PR #395, opened 2023-07-12
IOSS: Refactor/generalization of element many block parallel data read

tokusanya:
Cleaned up the interface using composition instead of templating. The generalization is mostly in terms of the ability to specify a subset of an entity grouping (Ioss::ElementBlock)

The current design should allow extension to Ioss:: ElementBlock, NodeSet and SideSet for when it is needed.


packages/seacas/libraries/ioss/src/Ioss_DatabaseIO.C
```diff
@@ -1362,17 +1361,25 @@ namespace Ioss {
...
+    size_t num_container = entity_container.size();
 
-    std::vector<size_t> offset = get_all_block_offsets(field_name, elem_blocks);
+    std::vector<size_t> offset = get_entity_offsets(field_name, entity_container);
 
-    for (size_t i = 0; i < num_blocks; i++) {
+    for(size_t i=0; i<num_container; i++) {
 
-      Ioss::ElementBlock *entity = elem_blocks[i];
+      auto entity = entity_container[i];
```
gsjaardema, line 1382: This should probably be `const auto *entity = entity_container[i];`

packages/seacas/libraries/ioss/src/Ioss_DatabaseIO.C
```diff
@@ -1400,8 +1408,8 @@ namespace Ioss {
         }
 
         size_t block_data_offset = offset[i] * field_byte_size;
-        auto   retval =
-            get_field_internal(entity, field, (char *)data + block_data_offset, block_data_size);
+
+        auto retval = get_field_internal(entity, field, (char*)data + block_data_offset, block_data_size);
```
gsjaardema, line 1412: Is the `block_data_size` accounting for offsetting into the `data` by `block_data_offset`.  That is, does `block_data_size` have `block_data_offset` subtracted from it somewhere before this use.

packages/seacas/libraries/ioss/src/Ioss_DatabaseIO.h
```diff
@@ -568,8 +569,9 @@ namespace Ioss {
       }
     }
 
-    virtual std::vector<size_t> get_all_block_field_data(const std::string &field_name, void *data,
-                                                         size_t data_size) const;
+    virtual std::vector<size_t> get_entity_field_data(const std::string &field_name,
+                                                      const std::vector<Ioss::ElementBlock*>& elem_blocks,
+                                                      void *data, size_t data_size) const;
```
gsjaardema, line 574: Did you plan to use `ElementBlockContainer` here?  I see the typedef at line 57.

tokusanya: That typedef is probably no longer needed ... I can remove it if you want

packages/seacas/libraries/ioss/src/Ioss_DatabaseIO.C
tokusanya: Yes, that is correct

packages/seacas/libraries/ioss/src/Ioss_DatabaseIO.C
tokusanya: I can change that too if you want

packages/seacas/libraries/ioss/src/Ioss_Decomposition.C
```diff
@@ -158,106 +158,110 @@ namespace Ioss {
...
+                                                       const std::vector<int>& blockSubsetFieldComponentCount) const
   {
     size_t count = 0;
 
-    for (size_t blk_seq = 0; blk_seq < blocks.size(); blk_seq++) {
-      const Ioss::BlockDecompositionData &block = blocks[blk_seq];
-      // Determine total number of ioss decomp entries based on field component count per block.
-      count +=
-          block_field_component_count[blk_seq] * (block.importMap.size() + block.localMap.size());
+    for(size_t i = 0; i < blockSubsetIndex.size(); i++) {
```
gsjaardema, line 179: In the loop starting at 166, you loop over `blockSubsetIndex`, but here you do a loop on the size of blockSubsetIndex, get the `blk_seq` from that and then the block from `m_data`. Is there a reason for the extra indirection here?

packages/seacas/libraries/ioss/src/Ioss_Decomposition.h
```diff
@@ -133,32 +133,44 @@ namespace Ioss {
...
-  IOSS_EXPORT size_t
-  get_all_block_ioss_element_size(const std::vector<BlockDecompositionData> &blocks);
+  class IOSS_EXPORT ElementBlockBatchOffset
+  {
+  public:
+    ElementBlockBatchOffset(const std::vector<BlockDecompositionData> &data)
+    : m_data(data) {}
+
+    ElementBlockBatchOffset()                                = delete;
+    ElementBlockBatchOffset(const ElementBlockBatchOffset &) = delete;
```
gsjaardema, line 143: Why is this deleted, but move constructor is defaulted?

packages/seacas/libraries/ioss/src/Ioss_Decomposition.h
```diff
@@ -133,32 +133,44 @@ namespace Ioss {
...
 
-  IOSS_EXPORT size_t
-  get_all_block_ioss_offset_size(const std::vector<BlockDecompositionData> &blocks,
-                                 const std::vector<int>                    &block_component_count);
+    ~ElementBlockBatchOffset() = default;
 
-  IOSS_EXPORT std::vector<size_t>
-              get_all_block_ioss_offset(const std::vector<BlockDecompositionData> &blocks,
-                                        const std::vector<int>                    &block_component_count);
+    size_t get_ioss_element_size(const std::vector<int64_t>& blockSubsetIndex) const;
```
gsjaardema, line 148: These probably need `IOSS_EXPORT` so they will work correctly on a windows build.

packages/seacas/libraries/ioss/src/Ioss_Decomposition.h
```diff
@@ -599,59 +616,60 @@ namespace Ioss {
...
+        int64_t blk_seq = blockSubsetIndex[bsi];
+        const Ioss::BlockDecompositionData& block = blocks[blk_seq];
+        size_t comp_count = blockComponentCount[bsi];
 
         for (size_t i = 0; i < block.localMap.size(); i++) {
           for (size_t j = 0; j < comp_count; j++) {
-            unsigned fileIndex = file_offset[blk_seq] + block.localMap[i] * comp_count + j;
-            unsigned iossIndex =
-                ioss_offset[blk_seq] + (i + block.localIossOffset) * comp_count + j;
+            unsigned fileIndex = fileOffset[bsi] + block.localMap[i] * comp_count + j;
```
gsjaardema, line 631: Is `unsigned` going to be large enough for big models or should `size_t` be used?

packages/seacas/libraries/ioss/src/Ioss_Decomposition.h
```diff
@@ -599,59 +616,60 @@ namespace Ioss {
...
+            unsigned iossIndex = iossOffset[bsi] + (i + block.localIossOffset) * comp_count + j;
             ioss_data[iossIndex] = file_data[fileIndex];
           }
         }
 
         for (size_t i = 0; i < block.importMap.size(); i++) {
           for (size_t j = 0; j < comp_count; j++) {
-            unsigned importIndex = import_offset[blk_seq] + i * comp_count + j;
+            unsigned importIndex = importOffset[bsi] + i * comp_count + j;
```
gsjaardema, line 639: `size_t` ?

packages/seacas/libraries/ioss/src/Ioss_Region.h
```diff
@@ -405,3 +409,80 @@ inline const std::vector<std::string> &Ioss::Region::get_qa_records() const
...
   return get_database()->get_qa_records();
 }
+
+namespace Ioss {
+
+template <typename T>
+bool verify_field_exists_in_entity_group(const std::string &field_name,
+                                         const std::vector<T*> &entity_container)
+{
+  bool found = false;
```
gsjaardema, line 419: This verifies that the field exists, but only verifies that the role is the same in a debug build.  Is it important whether the role is the same?  If so, then should be checked even in non-debug build.

packages/seacas/libraries/ioss/src/exodus/Ioex_DecompositionData.C
```diff
@@ -2174,6 +1934,249 @@ namespace Ioex {
...
+          int ierr = ex_get_partial_var(filePtr, step, EX_ELEM_BLOCK, var_index, id, offset + 1, count,
+                                        &file_data[file_index]);
+
+          if (ierr < 0) {
+            Ioex::exodus_error(filePtr, __LINE__, __func__, __FILE__);
+          }
+        }
+      }
+
+      decompData->m_decomposition.communicate_entity_data(file_data.data(), data,
```
gsjaardema, line 2174: I didn't think we needed to communicate for a "LINEAR" decompositions?  At least for element data...

Both parts of the if block look to be the same...

packages/seacas/libraries/ioss/src/unit_tests/unitMain.C
```diff
@@ -6,6 +6,8 @@
 
 #include <Ioss_CodeTypes.h>
 #include <gtest/gtest.h>
+#include <Ioss_CodeTypes.h>
```
gsjaardema, line 9: duplicate include

gsjaardema (commented):
Looks good, just a few questions where i didn't totally understand what was happening.

packages/seacas/libraries/ioss/src/Ioss_Decomposition.C
tokusanya: Yes, this is necessary because the blockSubsetFieldComponentCount vector needs to be referenced by an index `i` but that was not necessary in the for-loop on line 166

packages/seacas/libraries/ioss/src/Ioss_Decomposition.h
tokusanya: Ah, that was probably a typo ... the move constructor should probably be deleted because I want to use that class internally for now until I see how things shake up wih generalization to other GroupingEntity types

packages/seacas/libraries/ioss/src/Ioss_Decomposition.h
tokusanya: The class itself is already marked with `IOSS_EXPORT` ... do the individual functions in it also need to be marked?

packages/seacas/libraries/ioss/src/Ioss_Decomposition.h
tokusanya: You're right ... those are constructed from data that is `size_t` so they should be declared as `size_t`

packages/seacas/libraries/ioss/src/Ioss_Decomposition.h
tokusanya: Yup, same

packages/seacas/libraries/ioss/src/Ioss_Region.h
tokusanya: Yeah, it's important that they are the same so I'll make that check in all builds

packages/seacas/libraries/ioss/src/exodus/Ioex_DecompositionData.C
tokusanya: I was following the pattern that had been set up on 307 in Ioss_Decomposition.h for `communicate_element_data()` ..

packages/seacas/libraries/ioss/src/unit_tests/unitMain.C
tokusanya: I can remove the duplicate

packages/seacas/libraries/ioss/src/exodus/Ioex_DecompositionData.C
gsjaardema: I think you missed the `return` that is in the `LINEAR` part of the if block then...  There is also a copy in that block: (I removed asserts/comments just to shorten):
```
 if (m_method == "LINEAR") {
        auto size = localElementMap.size() * comp_count;
        std::copy(file_data, file_data + size, ioss_data);
        return;
      }
```

tokusanya: This function is the batch read version of `get_elem_var()` on line 1319 which does an exodus read from file into a data vector if the decomposition method is LINEAR. Since the generalization is based on subsetted element blocks, it is looping over those individual blocks and reading the data into the right locations just like `get_elem_var()`. I only mentioned the pattern above in `communicate_field_data()` as an example of where you check for the LINEAR decomposition method when doing parallel IO .. the batch equivalent of `communicate_field_data()` is `communicate_entity_data()` on line 653 in Ioss_Decomposition.h which does the early return that you mentioned.

gsjaardema: OK, I guess I got my brace count off and it looked like both branches of the if closed right before the `communicate_entity_data`.  I can now see that the LINEAR branch does not have the communicate in it.
Sorry for the false alarm.

gsjaardema: I don't know whether it would be worth pulling all the duplicate code into a function and calling that from each branch of the if.  Looks like only difference is the use of `data` in the linear portion and `file_data` in the other branch...

tokusanya: @gsjaardema I have another PR up that refactors and extracts that function

packages/seacas/libraries/ioss/src/exodus/Ioex_DecompositionData.C
```diff
@@ -2103,75 +2103,78 @@ namespace Ioex {
...
-          int ierr = ex_get_partial_var(filePtr, step, EX_ELEM_BLOCK, var_index, id, offset + 1, count,
-                                        &file_data[file_index]);
+    double *data = reinterpret_cast<double *>(iossData);
 
-          if (ierr < 0) {
-            Ioex::exodus_error(filePtr, __LINE__, __func__, __FILE__);
-          }
-        }
-      }
+    std::vector<int> blockComponentCount = get_block_component_count(blockSubsetIndex, blockFieldData);
```
gsjaardema, line 2164: Thanks, I think that is clearer and reduces code duplication.

gsjaardema:
Thanks, looks good.

PR #399, opened 2023-08-30
Aprepro enhanced loop

gsjaardema:
The loop command will now have an "automatic index" which is initialized to zero at the start of the loop and incremented by 1 each time through the loop.  The current name of the index is `__loop_#` where `#` is the 1-based nesting of the loop.
```
{loop(8)}
 This is the {__loop_1}'th iteration of the loop
{endloop}
```


gsjaardema:
Note to self -- These changes do not include changes to aprepro.ll -- the generated apr_scanner.cc was modified directly instead of modifying aprepro.ll and regenerating. (Currently can only regenerate with flex-2.5.39 which is not running on my laptop...)

gsjaardema:
> Note to self -- These changes do not include changes to aprepro.ll -- the generated apr_scanner.cc was modified directly instead of modifying aprepro.ll and regenerating. (Currently can only regenerate with flex-2.5.39 which is not running on my laptop...)

This has now been addressed.  all is ok with source and generated file in sync.

gsjaardema:
This has been run on all Sierra regression tests and results in no failures so it looks like there is no difference in behavior for backward compatibility.

PR #404, opened 2023-09-25
APREPRO: Make symrec definition available in aprepro.h

gsjaardema:


tjfulle:
@gsjaardema This looks great, thanks!!!

PR #407, opened 2023-10-03
Fixed build errors related to sys/statfs.h include

seanm:
The sys/statfs.h header is a linux thing, but it was being included for all non-Window/Mac. This broke building on FreeBSD, and probably other OSes too.

After consulting various man pages from various OSes, we now include the right headers per-OS for the statfs() function.


seanm:

PR #410, opened 2023-10-16
Fixed reported regression building with Emscripten

seanm:
#include <sys/statfs.h> not only for linux, but as a general fallback.

Thanks to Adrien Stucky for finding the problem.


mathstuf (approved):

seanm:

PR #411, opened 2023-10-16
Fix NFS detection on the BSDs

seanm:
- on the BSDs, look at the f_fstypename field to determine file system type, as this is the documented way to do that. The sqlite codebase also does so for macOS notably.
- removed dead code in the Windows case


mathstuf (approved):
Seems fine. Though I'll note that it looks like Windows does support NFS: https://learn.microsoft.com/en-us/windows-server/storage/nfs/nfs-overview

Something for the future where someone cares enough about it to add detection here.

seanm:

packages/seacas/libraries/ioss/src/Ioss_FileInfo.C
```diff
@@ -196,10 +193,15 @@ namespace Ioss {
         IOSS_ERROR(errmsg);
       }
       free(path);
-      return (stat_fs.f_type == NFS_FS);
+
+#if defined(__APPLE__) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__)
+      return (0 == ::strcmp("nfs", stat_fs.f_fstypename));
+#else
+      /* linux statfs defines that 0x6969 is NFS filesystem */
+      return (stat_fs.f_type == 0x6969);
```
gsjaardema, line 201: I would prefer that the `#define NFS_FS` be used here.  Many static analyzers complain about magic numbers.  I realize that the define would be used in exactly one place and that there are many other magic numbers throughout the code base, but ...

gsjaardema:
> Seems fine. Though I'll note that it looks like Windows does support NFS: https://learn.microsoft.com/en-us/windows-server/storage/nfs/nfs-overview
> 
> Something for the future where someone cares enough about it to add detection here.

Agreed.   There seem to be many more uses on Windows systems, so code should be added instead of just "punting..."

PR #430, opened 2024-01-16
TextMesh: Enable block and assembly omission/inclusion

tokusanya:
The text mesh database did not support omitted element blocks and assemblies . This is a fix to enable that oversight

### DIFF
--- a/packages/seacas/libraries/ioss/src/Ioss_DatabaseIO.C
+++ b/packages/seacas/libraries/ioss/src/Ioss_DatabaseIO.C
@@ -887,6 +887,24 @@ namespace Ioss {
       IOSS_ERROR(errmsg);
     }
 
+    if (!assemblyOmissions.empty() && !inclusions.empty()) {
+      std::ostringstream errmsg;
+      fmt::print(errmsg,
+                 "ERROR: Only one of element block inclusion or assembly omission can be non-empty"
+                 "       [{}]\n",
+                 get_filename());
+      IOSS_ERROR(errmsg);
+    }
+
+    if (!assemblyInclusions.empty() && !omissions.empty()) {
+      std::ostringstream errmsg;
+      fmt::print(errmsg,
+                 "ERROR: Only one of element block omission or assembly inclusion can be non-empty"
+                 "       [{}]\n",
+                 get_filename());
+      IOSS_ERROR(errmsg);
+    }
+
     if (!omissions.empty()) {
       blockOmissions.assign(omissions.cbegin(), omissions.cend());
       Ioss::sort(blockOmissions.begin(), blockOmissions.end());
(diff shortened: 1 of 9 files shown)


packages/seacas/libraries/ioss/src/text_mesh/Iotm_DatabaseIO.C
```diff
@@ -643,6 +643,23 @@ namespace Iotm {
     }
   }
 
+  void DatabaseIO::update_block_omissions_from_assemblies()
+  {
+    Ioss::SerializeIO serializeIO_(this);
```
gsjaardema, line 648: TextMesh should not need a call to Ioss::SerializeIO since it doesn't do IO ...

packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMesh.C
```diff
@@ -634,7 +707,7 @@ namespace Iotm {
     return assembly->id;
   }
 
-  Ioss::EntityType TextMesh::assembly_type_to_entity_type(const AssemblyType type) const
+  Ioss::EntityType TextMesh::assembly_type_to_entity_type(const AssemblyType type)
```
gsjaardema, line 710: Is there a reason the const was removed?

packages/seacas/libraries/ioss/src/text_mesh/Iotm_DatabaseIO.C
tokusanya: Ok, I'll remove this

tokusanya:
@gdsjaar Could you review the change?

packages/seacas/libraries/ioss/src/text_mesh/Iotm_DatabaseIO.C
```diff
@@ -643,6 +643,21 @@ namespace Iotm {
     }
   }
 
+  void DatabaseIO::update_block_omissions_from_assemblies()
+  {
+    if (!assemblyOmissions.empty()) {
+      assert(blockInclusions.empty());
```
gsjaardema, line 649: Is this the only place that we verify that if assemblyOmissions is not empty then blockInclusions must be empty?  If so, then this will not be checked in a non-debug build and will also not provide any error message in debug that a user could figure out what is happening...

packages/seacas/libraries/ioss/src/text_mesh/Iotm_DatabaseIO.C
```diff
@@ -643,6 +643,21 @@ namespace Iotm {
...
+    }
+
+    if (!assemblyInclusions.empty()) {
+      assert(blockOmissions.empty());
+    }
```
gsjaardema, line 654: Same as above.

gsjaardema (changes requested):
I think it would be good to have an error or warning message if a client specifies an invalid set of assembly/block omissions/inclusions instead of just an assert.

packages/seacas/libraries/ioss/src/Ioss_DatabaseIO.C
```diff
@@ -877,6 +877,16 @@ namespace Ioss {
   void DatabaseIO::set_block_omissions(const std::vector<std::string> &omissions,
                                        const std::vector<std::string> &inclusions)
   {
+    if (!omissions.empty() && !inclusions.empty()) {
```
gsjaardema, line 880: I don't think this is the same as what the asserts were checking.  The assert was assembly incluion and block omission or assembly omission and block inclusion.

This here is block omission and inclusion...

gsjaardema (approved):

gsjaardema (approved):
Looks good.  Thanks.

PR #442, opened 2024-03-07
Io shell omit blocks

gsjaardema:
Add the capbility to subset blocks  (and later nodesets, sidesets) in io_shell.

Possibly support selection of variables also...


gsjaardema:
Comitted a different way.  Abandoning this PR.

PR #445, opened 2024-03-14
IOSS: Extend Text mesh for edgeset

tokusanya:
Edgesets in TextMesh


tokusanya:
@gdsjaar There's probably some work to clean this up .... not sure why I had so many merge issues.

gsjaardema:
I think there is some confusion as to what an edgeblock and an edgeset are.  This PR seems to treat an edgeset as similar to a sideset.  An ioss sideset (surface) has one or more sideblocks and corresponds to an exodus sideset.   The ioss sideset/surface reads an exodus sideset and splits it into SideBlocks which are usually groups of sides with homogenous topology...

An edgeset (ioss and exodus) is a group of edges in the same way that a nodeset is a group of nodes and a elementset is a group of elements.  An edgeset references explicitly defined edges in the model which are defined in one or more edgeblocks -- Note that an edgeset does not contain edgeblocks, it groups a subset of the edges which are defined in the edgeblocks in a model.  The edgeblocks are similar to element blocks in that they are all of the same topology and are defined by their nodal connectivity.    Similarly faceblocks are all of the same topology and are defined by either their nodal connectivity or their edge connectivity.   If a model has edges and faces, then there is also a connectivity relationship from elements to faces to edges.   

See [ExodusII-Addendum](https://sandialabs.github.io/seacas-docs/ExodusII-Addendum.pdf) for a description of how the edge/face blocks/sets are defined and added to Exodus data model.  Chapter 2 has the description and Chapter 4 has the Exodus API modifications/extensions made to support these entity types. 

I think that what is being attempted here is to have some way of specifying that a sideset is referiring to the "second level" of its boundaries (edges) instead of the more common "first level" (faces) without using the current kluge. 

Let's continue this in a discussion...  #446

PR #447, opened 2024-03-29
IOSS: Add guard for ARM/ARM64-specific include on Mac

cticenhour:
Hi @gsjaardema! 👋🏻 

I ran across an error building SEACAS / IOSS on Intel Mac (the Mac SDK was missing `mach/arm/kern_return.h`) that I hadn't on my Apple Silicon system. Guarding against using that include on Intel Mac platforms seemed to be the right fix, so I wanted to contribute it back to you. 

EDIT: Double-checking with INL the terms of the CLA. Thanks for your patience!


gsjaardema:
I don't think that the CLA is needed in this case.  Thanks for the update.

PR #456, opened 2024-05-16
Add support for Exodus IOSS properties related to field interpretation to the Catalyst IOSS Database

ajpelle:
Added support (and appropriate tests) for Exodus properties relating to field interpretation to the Catalyst IOSS database. The properties added are:
- ENABLE_FIELD_RECOGNITION
- IGNORE_REALN_FIELDS
- FIELD_SUFFIX_SEPARATOR
- FIELD_STRIP_TRAILING_UNDERSCORE
- SURFACE_SPLIT_TYPE

The first four were implemented leveraging the get_fields() method in Ioss_Utils, while for SURFACE_SPLIT_TYPE we return empty sidesets and sideblocks (and -1 for related fields) if the SURFACE_SPLIT_TYPE changed between the write and read Catalyst database.


packages/seacas/libraries/ioss/src/catalyst/Iocatalyst_DatabaseIO.C
```diff
@@ -808,8 +816,25 @@ namespace Iocatalyst {
     }
 
     template <typename GroupingEntityT>
-    bool readFields(const conduit_cpp::Node &&parent, GroupingEntityT *block) const
+    bool readFields(const conduit_cpp::Node &&parent, GroupingEntityT *block, Ioss::DatabaseIO  *dbase) const
```
gsjaardema, line 819: Could also get database from `block->get_database()` so wouldn't necessarily need this argument.

packages/seacas/libraries/ioss/src/catalyst/Iocatalyst_DatabaseIO.C
```diff
@@ -846,7 +897,7 @@ namespace Iocatalyst {
       this->readProperties(child[detail::PROPERTIES], region);
 
       // read fields (meta-data only)
-      this->readFields(child[detail::FIELDS], region);
+      this->readFields(child[detail::FIELDS], region, region->get_database());
```
gsjaardema, line 900: Similar to above comment, could get this argument internal to the `readFields` function instead of passing it in here.  Would involve same function call, but just inside function instead of at each call site.

gsjaardema (approved):
LGTM.  A couple comments, but just FYI and not mandatory.

packages/seacas/libraries/ioss/src/catalyst/Iocatalyst_DatabaseIO.C
ajpelle: Good suggestion, I'll implement those changes and commit to this pull request!

ajpelle: Done; rdy to merge

PR #460, opened 2024-06-11
Correct Jamfile to match Seacas's CMake Library Names

akimler:
Jamfile change for Sierra's build system to support a spack built version of Seacas. 

This change correctly matches Seacas library names to CMake's version of Seacas's libraries and changes the path name to 'seacas' rather than 'sierra_seacas'

### DIFF
--- a/packages/seacas/Jamfile
+++ b/packages/seacas/Jamfile
@@ -138,7 +138,6 @@ local installed-developer-files =
 explicit install-serial-targets ;
 alias install-serial-targets ;
 
-
 # Dependencies listed in this target are installed in the developer's project.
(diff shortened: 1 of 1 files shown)


packages/seacas/Jamfile
```diff
@@ -174,7 +174,7 @@ install install-user-include0
 explicit install-user-include1 ;
 install install-user-include1
   : [ glob $(seacas-root)/libraries/suplib_c/*.h ]
-  : <location>$(install-root)/sierra_seacas/include
+  : <location>$(install-root)/seacas/include
```
gsjaardema, line 177: Could these be `$(seacas-root)/include` similar to the libraries?

PR #463, opened 2024-06-18
IOSS: Catalyst API 2

tjotaha:
Added support to database for using my_processor and processor_count IOSS properties when operating in
serial parallel mode and reading Conduit files.


gsjaardema (approved):

gsjaardema:
@tjotaha It looks like there is one failure and it is related to Catalyst.

tjotaha:
@gsjaardema I fixed the test failure on the Catalyst parallel build dashboard. I see a new failure on the Intel OneAPI dashboard due to the Fortran compiler (ifx) not begin found. Does something need to be adjusted on this machine?

PR #465, opened 2024-06-24
Fix export symbols

spyridon97:
Fix export symbols and comments section.


spyridon97:
@gdsjaar would you mind quickly approving this?

spyridon97:
FYI @tjotaha

PR #476, opened 2024-08-02
Dynamic topology phase 2

tokusanya:
IOSS: Add broker class to dynamic topology
    
    The original philosophy of one observer per region is still maintained
    with the observers that represent the same FEM model being grouped
    together. This means that topology changes on one observer is
    propagated to others in the same group

### DIFF
--- a/packages/seacas/libraries/ioss/src/exodus/Ioex_DatabaseIO.C
+++ b/packages/seacas/libraries/ioss/src/exodus/Ioex_DatabaseIO.C
@@ -171,6 +171,9 @@ namespace Ioex {
         IOSS_ERROR(errmsg);
       }
     }
+
+    open_root_group();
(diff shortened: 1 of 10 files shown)


packages/seacas/libraries/ioss/src/exodus/Ioex_DatabaseIO.C
```diff
@@ -171,6 +171,9 @@ namespace Ioex {
         IOSS_ERROR(errmsg);
       }
     }
+
+    open_root_group();
```
gsjaardema, line 175: These probably need to be the `open_root_group_nl()` and `open_child_group_nl(0)` since I think we are already in a mutex lock state here.

gsjaardema (approved):

gsjaardema (approved):

PR #481, opened 2024-08-15
Dynamic topology phase IV

tokusanya:
Functionality for reduction operations on groups


packages/seacas/libraries/ioss/src/exodus/Ioex_DatabaseIO.C
```diff
@@ -641,74 +641,85 @@ namespace Ioex {
...
         Ioss::Region *this_region = get_region();
+        int numSteps = 0;
         for (int i = 0; i < max_step; i++) {
           if (tsteps[i] <= max_time) {
-            this_region->add_state_nl(tsteps[i] * timeScaleFactor);
+            if(setRegionTimeSteps) {
+              this_region->add_state_nl(tsteps[i] * timeScaleFactor);
+            }
+
+            tsteps[i] *= timeScaleFactor;
```
tokusanya, line 686: @gsjaardema It is unclear if to return the original time or the scaled time ...

packages/seacas/libraries/ioss/src/Ioss_Region.C
```diff
@@ -308,6 +308,56 @@ namespace {
...
+    }
+  }
+
+  void locate_state(Ioss::DatabaseIO *db, const double targetTime, NearestGroupState& loc)
+  {
+    // Get state count and all states...
+    std::vector<double> timesteps = db->get_db_step_times();
+    int stepCount = timesteps.size();
+
+    if(targetTime < 0.0) {
```
gsjaardema, line 344: Is `targetTime < 0` a flag for different behavior?  If so, what happens if the user has negative timesteps on the database and wants to restart at a negative time

packages/seacas/libraries/ioss/src/Ioss_Region.C
```diff
@@ -308,6 +308,56 @@ namespace {
...
+
+  struct NearestGroupState
+  {
+    std::string group{"/"};
+    int nearestState{-1};
+    double nearestTime{-std::numeric_limits<double>::max()};
+  };
+
+  void get_state_time_round_down(Ioss::DatabaseIO *db, const double targetTime, NearestGroupState& loc)
+  {
```
gsjaardema, line 320: This function looks almost exactly like the `locate_state` function below.  except for the part where the function below calls this function...   Do we need both, or at least combine similar parts.

packages/seacas/libraries/ioss/src/Ioss_Region.C
```diff
@@ -308,6 +308,56 @@ namespace {
...
+    int stepCount = timesteps.size();
+
+    if(targetTime < 0.0) {
+      get_state_time_round_down(db, targetTime, loc);
+    }
+    else {
+      double minTimeDiff = loc.nearestState < 0 ? std::numeric_limits<double>::max() : std::fabs(targetTime - loc.nearestTime);
+      for(int istep = 1; istep <= stepCount; istep++) {
+        double stateTime = timesteps[istep-1];
+        double stepTimeDiff = std::fabs(targetTime - stateTime);
```
gsjaardema, line 351: Is there a reason why `std::fabs` is used here, but `std::abs` is used on line 328 above.

packages/seacas/libraries/ioss/src/Ioss_Region.C
```diff
@@ -3055,4 +3105,28 @@ namespace Ioss {
       topologyObserver->reset_topology_modification();
     }
   }
+
+  std::string Region::get_group_name() const
```
gsjaardema, line 3109: Is this needed by external clients...   Not sure it makes sense to expose the concept of a "group" to external clients since it is an artifact of how we are using exodus.   Is it possible to isolate groups at the exodus level.  If not, is there some more descriptive name that we can use at the higher levels...

packages/seacas/libraries/ioss/src/Ioss_Region.C
```diff
@@ -3055,4 +3105,28 @@ namespace Ioss {
...
+  {
+    return get_database()->get_group_name();
+  }
+
+  std::tuple<std::string, int, double> Region::locate_db_state(const double targetTime)
```
gsjaardema, line 3114: Does this only work for the cases where the exodus file contains groups, or will it return the "group" correspoinding to a set of files `-s000X` storing topology changed data...

packages/seacas/libraries/ioss/src/Ioss_Region.h
```diff
@@ -313,6 +314,10 @@ namespace Ioss {
     bool load_group_mesh(const std::string &child_group_name);
     bool load_group_mesh(const int child_group_index);
 
+    std::string get_group_name() const;
+
+    IOSS_NODISCARD std::tuple<std::string, int, double> locate_db_state(const double targetTime);
```
gsjaardema, line 319: Some compilers/analyzers will warn/complain about value parameters with `const` since it doesn't matter if it is const or not at this level since a copy is being passed and not the address of the clients value.

packages/seacas/libraries/ioss/src/Ioss_DatabaseIO.h
```diff
@@ -356,7 +358,13 @@ namespace Ioss {
     void get_step_times()
     {
       IOSS_FUNC_ENTER(m_);
-      return get_step_times_nl();
+      get_step_times_nl();
+    }
+
+    std::vector<double> get_db_step_times()
```
gsjaardema, line 364: I know you don't like comments, but how does a client know whether to use this function or the `get_step_times()` function?

packages/seacas/libraries/ioss/src/exodus/Ioex_DatabaseIO.C
```diff
@@ -641,74 +641,85 @@ namespace Ioex {
...
 
-  void DatabaseIO::get_step_times_nl()
+  std::vector<double> DatabaseIO::internal_get_step_times_nl(bool setRegionTimeSteps)
   {
     bool                exists         = false;
     double              last_time      = DBL_MAX;
+    int                 tstepCount     = 0;
     std::vector<double> tsteps(0);
 
+    int& timestepCount = setRegionTimeSteps ? m_timestepCount : tstepCount;
```
gsjaardema, line 651: This is a little obscure since it is optionally setting the member variable.  A comment might be useful since it looks very strange why you have a reference here...

packages/seacas/libraries/ioss/src/exodus/Ioex_DatabaseIO.C
gsjaardema: I think scaled time.

packages/seacas/libraries/ioss/src/gen_struc/Iogs_DatabaseIO.C
```diff
@@ -489,6 +489,18 @@ namespace Iogs {
     }
   }
 
+  std::vector<double> DatabaseIO::get_db_step_times_nl()
+  {
+    std::vector<double> timesteps;
+
+    int time_step_count = m_generatedMesh->timestep_count();
+    for (int i = 0; i < time_step_count; i++) {
```
gsjaardema, line 497: After line 496: `timesteps.reserve(time_step_count);` to avoid resizing/reallocations

packages/seacas/libraries/ioss/src/generated/Iogn_DatabaseIO.C
```diff
@@ -590,6 +590,18 @@ namespace Iogn {
     }
   }
 
+  std::vector<double> DatabaseIO::get_db_step_times_nl()
+  {
+    std::vector<double> timesteps;
+
+    int time_step_count = m_generatedMesh->timestep_count();
+    for (int i = 0; i < time_step_count; i++) {
```
gsjaardema, line 598: After line 597: `timesteps.reserve(time_step_count)`

packages/seacas/libraries/ioss/src/text_mesh/Iotm_DatabaseIO.C
```diff
@@ -643,6 +643,18 @@ namespace Iotm {
     }
   }
 
+  std::vector<double> DatabaseIO::get_db_step_times_nl()
+  {
+    std::vector<double> timesteps;
+
+    int time_step_count = m_textMesh->timestep_count();
+    for (int i = 0; i < time_step_count; i++) {
```
gsjaardema, line 651: Same comment as above -- reserve

packages/seacas/libraries/ioss/src/unit_tests/UnitTestDynamicTopology.C
```diff
@@ -1106,6 +1108,57 @@ TEST(TestDynamicRead, single_file_simple_topology_modification)
   cleanup_single_file(outFile);
 }
 
+std::tuple<std::string, int, double> read_and_locate_db_state(const OutputParams& params, const double targetTime)
```
gsjaardema, line 1111: Might be good to add more tests -- one with all negative times, one with negative and positive.  Also have times in random order `5, 1, 4` to simulate the cyclic time option...

gsjaardema (changes requested):
Looks good.  A few questions/comments/suggestions.

packages/seacas/libraries/ioss/src/Ioss_Region.C
tokusanya: This is based on Frio Restart code that I assume needed the distinction between negative and positive start times in terms of locating the right state .... I could add unit tests with negative times to see how it affects the results

packages/seacas/libraries/ioss/src/Ioss_Region.C
tokusanya: See comment above

packages/seacas/libraries/ioss/src/Ioss_Region.C
tokusanya: None ... it is what is being used in Frio_Restart so wasn't sure if the distinction was needed. Should I make them consistent?

packages/seacas/libraries/ioss/src/Ioss_Region.C
tokusanya: It probably doesn't need to exposed externally .. I'll make it a private/function.

packages/seacas/libraries/ioss/src/Ioss_Region.C
tokusanya: Yeah, this is really only for group files but I might be able to modify it to scan the multi-file case....the return string would have to be the database name, I guess

packages/seacas/libraries/ioss/src/Ioss_Region.h
tokusanya: Ok, I can change that programming style then ... I'm used to the idea of making sure that all possible arguments that *can* be const *should* be const

packages/seacas/libraries/ioss/src/Ioss_DatabaseIO.h
tokusanya: I'll add the necessary comments

packages/seacas/libraries/ioss/src/exodus/Ioex_DatabaseIO.C
tokusanya: Yeah, the reference is to make sure that no Region modifications occur based on the input flag but I can add a comment for that

packages/seacas/libraries/ioss/src/gen_struc/Iogs_DatabaseIO.C
tokusanya: Ok, will do

packages/seacas/libraries/ioss/src/generated/Iogn_DatabaseIO.C
tokusanya: Will do

packages/seacas/libraries/ioss/src/text_mesh/Iotm_DatabaseIO.C
tokusanya: Will do

packages/seacas/libraries/ioss/src/unit_tests/UnitTestDynamicTopology.C
tokusanya: Yup, will do

packages/seacas/libraries/ioss/src/Ioss_Region.C
tokusanya: So it appears that the difference is the monotonicity check ... it's slightly different for negative times and positive times

packages/seacas/libraries/ioss/src/Ioss_Region.C
gsjaardema: Yes, probably.

packages/seacas/libraries/ioss/src/Ioss_Region.C
gsjaardema: I guess it depends on how complete and broad we are trying to make the capability.  I don't know what the clients will expect -- I could see them wanting to use the same interface for all database types (that support restart/output) whether they use groups or different files, or a different method all together....  If it doesn't look too difficult then might be worthwhile...

packages/seacas/libraries/ioss/src/Ioss_Region.C
gsjaardema: It looks like lines 324 .. 335 are exactly the same as lines 348 .. 359 unless I'm missing something...

tokusanya: Lines 329 and 352 are where the monotonicity checks are made ... one is a strict < and the other is <=

packages/seacas/libraries/ioss/src/Ioss_Region.C
tokusanya: Done

packages/seacas/libraries/ioss/src/unit_tests/UnitTestDynamicTopology.C
tokusanya: I thought the CYCLIC_FILE option only referred to the creation of the filenames based on a modulo reduction. How does it affect the timestep since I thought the timesteps associated with states always had to be monotone increasing. I typically see warning messages if the times are in random order

gsjaardema: It is posible to use the cyclic option on timesteps within a single file in which case the timesteps would not be monotonically increasing.  There is a warning, but it is allowed.   I think CYCLIC_FILE puts each timestep in a single file and then reuses those files, but there is also a way that a single file will be used and timesteps overwritten.  

For example, if cycle is set to 3 then there would only be 3 timesteps in the file and they would be 1, 2, 3.  Then 1 would be overwritten with 4, 2 with 5, 3 with 6,.  Then 4 with 7, ...

This is in Ioss_DatabaseIO.h:
```
    /*! Typically used for restart output, but can be used for all output...
     * Maximum number of states on the output file.  Overwrite the existing
     * steps in a cyclic manner once exceed this count.  Note that this breaks
     * the convention that times be monotonically increasing on an exodusII file.
     * Used by derived classes if they support this capability...
     */
    mutable int cycleCount{0};
```

packages/seacas/libraries/exodus/src/ex_inquire.c
```diff
@@ -975,6 +975,18 @@ static int ex_inquire_internal(int exoid, int req_info, int64_t *ret_int, float
 #endif
     break;
 
+  case EX_INQ_FILE_FORMAT:
```
gsjaardema, line 978: If this stays, then it should also distinguish between all the different formats.  classic, netcdf4-classic, netcdf4, netcd3-large-model, netcdf5.

I'm not sure it is needed though, see comment below...

packages/seacas/libraries/ioss/src/exodus/Ioex_BaseDatabaseIO.C
```diff
@@ -502,6 +502,24 @@ namespace Ioex {
     ex_set_max_name_length(m_exodusFilePtr, maximumNameLength);
   }
 
+  bool BaseDatabaseIO::supports_group_nl()
```
gsjaardema, line 505: I'm not sure this is the correct way to do this...  Basically, prior to creating the database, we want to know if the database type specified by the user supports groups -- exodus, cgns, adios, faodal, ...  Not whether a specific database file that we have already created.

So, we just have to know whether exodus was compiled with netcdf-4 capability enabled and then `EXODUS` output type would return true for supporting groups and we could then created a "grouped database".   

This would also then be a compile-time check instead of runtime and we would not need the locking and `_nl` types. 

I think checking after the database is created is too late and doesn't give the client the information that they want.  If they want to use groups and the chosen format supports groups, then they can create the database (and we make sure that netcdf-4 format is selected as we currently do).

packages/seacas/libraries/ioss/src/exodus/Ioex_DatabaseIO.C
```diff
@@ -648,6 +648,7 @@ namespace Ioex {
     int                 tstepCount     = 0;
     std::vector<double> tsteps(0);
 
+    // Use reference to make sure that no Region modifications occur based on the input flag
```
gsjaardema, line 651: I would say something like the `setRegionTimeSteps` flag determines whether we are actually populating the region timesteps or just querying the timesteps that are on a specific database without populating the regions timesteps data and setting the number of timesteps on the region... 

Maybe not much better than the coment above, but maybe says a little more about why  a client would want to do that...

packages/seacas/libraries/ioss/src/Ioss_DatabaseIO.h
```diff
@@ -246,6 +246,16 @@ namespace Ioss {
       flush_database_nl();
     }
 
+    /** \brief Checks if a database type supports groups
+     *
+     *  \returns True if successful.
+     */
+    bool supports_group()
```
gsjaardema, line 253: I think this is exposing the method that we are using in exodus through netCDF -- `groups` as a general concept in the IOSS API.  

I think that instead the client knows that their output/model is going to have topology changes and want to store that in the best way possible.  One way is that they know they are using exodus as the output format and they either want single file output or each change in a separate file.  They don't care about how this is done internally in the file.

I don't know what our API-visible interface for this should be, but `groups` is very overloaded term and could be confused with `assemblies`.  Or, to put another way, `groups` doesn't suggest to me any association with changing topology.  

It may be ok to use `group` down at the exodus level, but if cgns supports something like `delta-sets` as their method for topology changes then we have confusion in the API...  

I'm not sure what to suggest, but seeing all this group-related api functions at this level just seems wrong...

Maybe something like `change-set` or ?   -- All database type could/might support this.  In some it is a set of appropriately-named files; in some it uses the netCDF groups, if some other type each file contains full model topology and there is some metadata file that says what happened from file to file...

At creation, the client specifies that topology changes will happen, so IOSS does the appropriate action (whatever that might be).  On read, IOSS queries the database(s) and can report that there are topology changess or multiple "change-sets".

Note that the "change-sets" might all contain the same topology -- is this is a run with multiple restarts, each restart state might be in a separate "change-set" even though the topology from one to another is the same.

gsjaardema: I think we have the basics of this working in exodus at this point, but now need to plan for how the client interfaces with the concept and how the concept gets expanded out as a "database-agnostic" concept that encompasses and encapsulates as much similar behavior as possible. It also needs to plan for extension.

I don't know what that API view looks like, but I don't think "groups" is the correct   one.

gsjaardema (changes requested):

packages/seacas/libraries/ioss/src/Ioss_DatabaseIO.h
tokusanya: @gdsjaar I think we probably need to meet to discuss  these concepts and nail down what the API should eventually look like

packages/seacas/libraries/ioss/src/exodus/Ioex_BaseDatabaseIO.C
tokusanya: I get what you are saying here but it appears that we may have different interpretations of what this function is supposed to do. I think you are right that the `change-set` concept that is handled with groups in exodus is spilling out and needs to be made database-agnostic. 

Based on the usage of this function in the unit tests, my goal was to find a way to do reduction operations on the databases depending on whether an input file was written in a format that supports groups (as opposed to knowing if the currently compiled database format *can* support groups). If the input file format doesn't support groups (or `change-sets` as we are now referring to them), then we do reductions across a set of files which may be `linear` or `cyclic` , otherwise we do the reduction across the `change-sets` within the file.

I would really like to meet up to have a more in-depth discussion on these concepts and API

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.C
```diff
@@ -36,14 +36,463 @@
...
+  }
+}
+
+void locate_state(Ioss::DatabaseIO *db, double targetTime, DatabaseState& loc)
+{
+  // Get state count and all states...
+  std::vector<double> timesteps = db->get_db_step_times();
+  int stepCount = timesteps.size();
+
+  if(targetTime < 0.0) {
```
gsjaardema, line 91: This is much clearer than previous code.

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.C
```diff
@@ -36,14 +36,463 @@
...
+                 const std::string &filename,
+                 const std::string &db_type,
+                 Ioss::DatabaseUsage db_usage)
+{
+  bool exists = false;
+  int par_size = util.parallel_size();
+  int par_rank = util.parallel_rank();
+  bool is_parallel = par_size > 1;
+  std::string full_filename = filename;
+  if (is_parallel && db_type == "exodusII" && db_usage != Ioss::WRITE_HISTORY) {
```
gsjaardema, line 128: Maybe cgns also...

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.C
```diff
@@ -36,14 +36,463 @@
...
+int file_exists( const Ioss::ParallelUtils &util,
+                 const std::string& filename,
+                 std::string& message,
+                 bool specifiedDecomp )
+{
+  std::string filenameBase = filename;
+  const int par_size = util.parallel_size();
+  const int par_rank = util.parallel_rank();
+
+  if( par_size > 1 && !specifiedDecomp ) {
```
gsjaardema, line 157: Is the `specifiedDecomp` meaning that the user specified a decomposition method which implies that reading is from a single file instead of file-per-rank?

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.C
```diff
@@ -36,14 +36,463 @@
...
+
+  Ioss::FileInfo file = Ioss::FileInfo(filenameBase);
+  return file.parallel_exists(util.communicator(), message);
+}
+
+bool internal_decomp_specified(const Ioss::PropertyManager& props)
+{
+  bool internalDecompSpecified = false;
+
+  const std::string restartDecompMethod("RESTART_DECOMPOSITION_METHOD");
```
gsjaardema, line 169: This could also be `DECOMPOSITION_METHOD`

But I'm not sure this is robust since we now have similar code in multiple places -- exodus IOFactory and here and maybe other places.

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.C
```diff
@@ -36,14 +36,463 @@
...
+    }
+    IOSS_ERROR(errmsg);
+  }
+
+  if( exists == par_size ) {
+    db = Ioss::IOFactory::create(db_type, filename, Ioss::QUERY_TIMESTEPS_ONLY,
+                                 util.communicator(), properties);
+    int bad_count = 0;
+    std::string error_message;
+    bool is_exodus_file = db != nullptr && db->ok(false, &error_message, &bad_count);
```
gsjaardema, line 221: Why does exodus appear here?  Can it be type agnostice?

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.C
```diff
@@ -36,14 +36,463 @@
...
+    }
+  }
+}
+
+void locate_linear_multi_db_state(Ioss::DatabaseIO* db, double targetTime,
+                                  const std::string& ioDB, const std::string& dbType,
+                                  DatabaseState& loc)
+{
+  auto util = db->util();
+  FileNameGenerator generator = [](const std::string& baseFileName, unsigned step) {
```
gsjaardema, line 275: Maybe I'm misreading something here, but isn't this casting the type of a lambda into a std::function (or whatever FileNameGenerator is)?

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.C
```diff
@@ -415,6 +842,47 @@ std::string DynamicTopologyFileControl::get_unique_filename(Ioss::DatabaseUsage
...
+{
+  std::string filename = baseFileName;
+
+  if(fileCyclicCount > 0)
+  {
+    // The file suffix cycles through the first fileCyclicCount'th entries in A,B,C,D,E,F,...
+    if(step == 0)
+      step++;
+
+    static std::string suffix = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
```
gsjaardema, line 865: Seems like we have this in multiple places...  Maybe one is for input and one for output...

packages/seacas/libraries/ioss/src/Ioss_Region.C
```diff
@@ -3055,4 +3042,55 @@ namespace Ioss {
...
+                                           ifDatabaseExists, dbChangeCount);
+
+    return fileControl.get_db_max_time();
+  }
+
+  std::tuple<std::string, int, double> Region::get_db_min_time() const
+  {
+    IOSS_FUNC_ENTER(m_);
+    auto db = get_database();
+    if (!db->is_input() && db->usage() != WRITE_RESULTS && db->usage() != WRITE_RESTART) {
```
gsjaardema, line 3083: I guess this is for heartbeat and history?

packages/seacas/libraries/ioss/src/exodus/Ioex_ParallelDatabaseIO.C
```diff
@@ -553,7 +556,7 @@ namespace Ioex {
 
 #if !defined(__IOSS_WINDOWS__)
     if (!path.empty()) {
-      chdir(current_cwd);
+      (void)chdir(current_cwd);
```
gsjaardema, line 559: That probably quiets the warning on some platforms, but I should probably actually do some error checking to make sure the chdir worked...

packages/seacas/libraries/ioss/src/Ioss_DatabaseIO.h
```diff
@@ -246,63 +246,75 @@ namespace Ioss {
       flush_database_nl();
     }
 
-    /** \brief If a database type supports groups and if the database
-     *         contains groups, open the specified group.
+    /** \brief If a database type supports change sets and if the database
```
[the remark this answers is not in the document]
tokusanya: Yes, I wanted `change_set` to be an IOSS concept that encapsulates the separation layer between it and the various formats. However, the separation I originally had in mind was the internal implementation of the "multiple-file-per-database" concept (which is done through `groups` in Exodus. Upon refactoring a lot of the functiionality in Ioss_DynamicTopology.C, I realized that `change_set` could be a more generalized concept and had actually started trying to write a `ChangeSet` class that would generalize the "group" file, linear multi-file and cyclic multi-file ideas. If that sounds like a good idea, then it nicely dovetails into what you mentioned above regarding knowing the database type for IOSS to choose what method to use. However, in a case like exodus which supports both group and multi file, you would need finer granularity to tell IOSS which one to use

packages/seacas/libraries/ioss/src/Ioss_DatabaseIO.h
```diff
@@ -246,63 +246,75 @@ namespace Ioss {
...
-      return open_child_group_nl(index);
+      return change_set_describe_nl(return_full_names);
     }
 
-    Ioss::NameList groups_describe(bool return_full_names = false)
+    /** \brief Checks if a database type supports changes sets
+     *
+     *  \returns True if successful.
+     */
+    bool supports_change_set()
```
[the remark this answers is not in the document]
tokusanya: As per my comment above, if we are generalizing `change_set` to mean internal and multi-files, then the meaning of this function changes. I initially wrote it to mean that the underlying format supports internal files but I can try to extend it to the more general idea that you proposed

packages/seacas/libraries/ioss/src/Ioss_DatabaseIO.h
```diff
@@ -790,12 +818,12 @@ namespace Ioss {
       return elemMap.global_to_local(global);
     }
 
-    virtual int  num_child_group_nl() { return 0; }
-    virtual bool open_child_group_nl(int /* index */) { return false; }
-    virtual bool open_root_group_nl() { return false; }
-    virtual bool open_group_nl(const std::string & /* group_name */) { return false; }
-    virtual bool create_subgroup_nl(const std::string & /* group_name */) { return false; }
-    virtual Ioss::NameList groups_describe_nl(bool /* return_full_names */)
+    virtual bool supports_change_set_nl() { return false; }
```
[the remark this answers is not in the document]
tokusanya: See above

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.C
tokusanya: The power of refactoring 😸

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.C
tokusanya: Done ... that comes from Framework IO

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.C
tokusanya: Yes, I believe the function that computes it is based on this

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.C
tokusanya: So should I check for both strings or have some encapsulation of this in Ioss::Utils ?

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.C
tokusanya: I'll remove references to exodus ... it's how it appeared in Framework IO 😬

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.C
tokusanya: I believe this is standard where you construct a std::function from a lambda. It's not a cast

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.C
tokusanya: Are you talking about line 865 or the function?

packages/seacas/libraries/ioss/src/Ioss_Region.C
tokusanya: I wonder if we can simplify that line for input only ... can an input file have usage of `WRITE_RESULTS` or `WRITE_RESTART` ?

packages/seacas/libraries/ioss/src/Ioss_DatabaseIO.h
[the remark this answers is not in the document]
gsjaardema: Yes, that is what I thought we had decided on when change-set was originally proposed:

> Maybe something like change-set or ? -- All database type could/might support this. In some it is a set of appropriately-named files; in some it uses the netCDF groups, if some other type each file contains full model topology and there is some metadata file that says what happened from file to file...

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.C
gsjaardema: The line is what caught my attention, but I think the same functionality of generating a cyclic file name is present in many locations; especially now that we have pulled some framework code down into IOSS.  Maybe now is time to consolidate into a single function if possible.

packages/seacas/libraries/ioss/src/Ioss_Region.C
gsjaardema: It is for non-input (i.e. output) and I think the only output tyeps are WRITE_{results|restart|history|heartbeat}
(packages/seacas/libraries/ioss/src/Ioss_DBUsage.h)

tokusanya: So yeah, this function is meant for RESULTS and RESTART, hence the early return ... not sure if to limit it to output (WRITE_*)  though

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.C
tokusanya: Ok, I'll look into consolidation

packages/seacas/libraries/ioss/src/Ioss_ChangeSet.h
```diff
@@ -0,0 +1,101 @@
...
+// NTESS, the U.S. Government retains certain rights in this software.
+//
+// See packages/seacas/LICENSE for details
+
+#pragma once
+
+#include "Ioss_DatabaseIO.h"
+#include "Ioss_DBUsage.h"
+#include "Ioss_Region.h"
+#include "Ioss_ParallelUtils.h"
```
gsjaardema, line 12: Probably add `PropertyManager.h` include since referenced in the file...   Or could forward-declare it and `Ioss::Region`...

packages/seacas/libraries/ioss/src/Ioss_ChangeSet.h
```diff
@@ -0,0 +1,101 @@
...
+#include <iomanip>
+#include <string>
+#include <functional>
+#include <map>
+
+namespace Ioss {
+
+using FileNameGenerator = std::function<std::string(const std::string& baseFileName, unsigned step)>;
+
+enum ChangeSetFormat {
```
gsjaardema, line 28: Maybe `enum class`.  

I've used that some where I can.  Since this is new code it should not break existing users.

packages/seacas/libraries/ioss/src/Ioss_ChangeSet.h
```diff
@@ -0,0 +1,101 @@
...
+  virtual void close_change_set(unsigned index);
+
+  IOSS_NODISCARD unsigned size() const { return m_changeSetNames.size(); }
+  IOSS_NODISCARD const std::vector<std::string>& names() const { return m_changeSetNames; }
+  IOSS_NODISCARD std::string get_change_set_name(unsigned index) const;
+
+  IOSS_NODISCARD unsigned get_file_cyclic_count() const { return m_fileCyclicCount; }
+
+private:
+  ChangeSet() = delete;
```
gsjaardema, line 74: If deleted, no reason to put down in private.  Put up by other constructors so easier to see what exists and what doesn't.

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.C
```diff
@@ -36,670 +37,882 @@
...
+    // on some of the larger machines, filesystems, and processor counts...
+    Ioss::FileInfo file = Ioss::FileInfo(full_filename);
+    exists = file.exists();
   }
 
-  void DynamicTopologyObserver::set_topology_modification_nl(unsigned int type)
-  {
-    m_topologyModification |= type;
-    m_cumulativeTopologyModification |= type;
+  if (is_parallel) {
```
gsjaardema, line 167: Is it better to use the `Ioss::FileInfo::parallel_exists` ?

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.C
```diff
@@ -36,670 +37,882 @@
...
+{
+  bool exists = false;
+  int par_size = util.parallel_size();
+  int par_rank = util.parallel_rank();
+  bool is_parallel = par_size > 1;
+  std::string full_filename = filename;
+  if (is_parallel &&
+      (db_type == "exodusII" || db_type == "cgns") &&
+      (db_usage != Ioss::WRITE_HISTORY)) {
+    full_filename = Ioss::Utils::decode_filename(filename, par_rank, par_size);
```
gsjaardema, line 156: Does this correctly handle the file-per-rank and auto-decomp cases?

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.C
```diff
@@ -36,670 +37,882 @@
...
+}
 
-      if (nullptr != m_notifier) {
-        for (auto observer : m_notifier->get_observers()) {
-          observer->set_topology_modification_nl(type);
-        }
+
+namespace Ioss {
+
+void DynamicTopologyObserver::check_region() const
```
gsjaardema, line 180: Some code above uses `verify_XXX` for a similar type function.  Maybe good to be consistent.
Also `check_region` or `verify_region` not very descriptive.

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.h
```diff
@@ -20,7 +20,7 @@
 
 #include <iosfwd> // for ostream
 #include <sstream>
-#include <string> // for string, operator<
+#include <string>  // for string, operator<
```
gsjaardema, line 23: Is there need for `iosfwd` and `sstream` includes anymore?

Also `strings`

packages/seacas/libraries/ioss/src/exodus/Ioex_ChangeSet.h
```diff
@@ -0,0 +1,69 @@
...
+//
+// See packages/seacas/LICENSE for details
+
+#pragma once
+
+#include "Ioss_CodeTypes.h"
+#include "Ioss_DBUsage.h"    // for DatabaseUsage
+#include "Ioss_ChangeSet.h"  // for ChangeSet
+#include "Ioss_ChangeSetFactory.h"  // for ChangeSetFactory
+#include "Ioss_Map.h"        // for Map
```
gsjaardema, line 13: Is `Ioss_Map.h` needed

packages/seacas/libraries/ioss/src/exodus/Ioex_ChangeSet.h
```diff
@@ -0,0 +1,69 @@
...
+#include <cstddef>           // for size_t
+#include <cstdint>           // for int64_t
```
gsjaardema, line 15: Might not need this.

packages/seacas/libraries/ioss/src/Ioss_Region.C
```diff
@@ -3053,4 +3040,55 @@ namespace Ioss {
...
   }
+
+  std::string Region::get_internal_change_set_name() const
+  {
+    return get_database()->get_internal_change_set_name();
+  }
+
+  std::tuple<std::string, int, double> Region::locate_db_state(double targetTime) const
+  {
+    IfDatabaseExistsBehavior  ifDatabaseExists{Ioss::DB_OVERWRITE};
```
gsjaardema, line 3051: Lines 3051 and 3052 should not be here.

packages/seacas/libraries/ioss/src/Ioss_Region.C
```diff
@@ -3053,4 +3040,55 @@ namespace Ioss {
...
+
+  std::tuple<std::string, int, double> Region::get_db_max_time() const
+  {
+    IOSS_FUNC_ENTER(m_);
+    auto db = get_database();
+    if (!db->is_input() && db->usage() != WRITE_RESULTS && db->usage() != WRITE_RESTART) {
+      return std::make_tuple(get_internal_change_set_name(), currentState, stateTimes[0]);
+    }
+
+    IfDatabaseExistsBehavior  ifDatabaseExists{Ioss::DB_OVERWRITE};
```
gsjaardema, line 3068: These two lines don't do anything...

gsjaardema: Or, you should pass `Ioss::DB_OVERWRITE` in place of `ifDatabaseExists` and `1` in place of `dbChangeCount` ?

Not totally sure what is meant here.  Could also use `get_topology_change_count()` and `get_if_database_exists_behavior()`...

packages/seacas/libraries/ioss/src/Ioss_ChangeSetFactory.h
```diff
@@ -0,0 +1,57 @@
...
+ */
+class IOSS_EXPORT ChangeSetFactory
+{
+public:
+  virtual ~ChangeSetFactory() = default;
+  IOSS_NODISCARD static std::shared_ptr<ChangeSet> create(Ioss::Region *region);
+
+  static int                        describe(NameList *names);
+  IOSS_NODISCARD static NameList    describe();
+  static void                       clean();
```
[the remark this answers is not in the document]
tokusanya: The ChangeSetFactory or those functions? A lot of the design for the factory is based on IOFactory that implements those functions

packages/seacas/libraries/ioss/src/Ioss_ChangeSet.C
```diff
@@ -0,0 +1,359 @@
...
+
+
+#include <assert.h>
+
+
+namespace {
+int file_exists( const Ioss::ParallelUtils &util,
+                 const std::string& filename,
+                 std::string& message,
+                 bool specifiedDecomp )
```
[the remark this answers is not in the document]
tokusanya: Yeah, the flag just serves to specify that whether the user is using a specified decomposition. It's computed from `internal_decomp_specified(const Ioss::PropertyManager& props)`

packages/seacas/libraries/ioss/src/Ioss_ChangeSetFactory.h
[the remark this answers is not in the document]
gsjaardema: `clean`

packages/seacas/libraries/ioss/src/Ioss_ChangeSet.C
```diff
@@ -0,0 +1,359 @@
...
+      get_linear_multi_file_change_sets();
+    }
+  }
+
+  void ChangeSet::verify_change_set_index(unsigned index) const
+  {
+    if(index >= m_changeSetNames.size()) {
+      std::ostringstream errmsg;
+      fmt::print(errmsg,
+                 "Invalid change set index {} with a max range of {}\n",
```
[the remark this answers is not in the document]
tokusanya: Yeah, I can change the string

packages/seacas/libraries/ioss/src/Ioss_ChangeSet.C
```diff
@@ -0,0 +1,359 @@
...
+      }
+    }
+
+    // open db
+    const std::string&  ioDB = m_changeSetNames[index];
+    const std::string dbType = m_dbType;
+
+    db = Ioss::IOFactory::create(dbType, ioDB, usage, util().communicator(),
+                                 get_database()->get_property_manager());
+    int bad_count = 0;
```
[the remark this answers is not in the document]
tokusanya: I can do that

packages/seacas/libraries/ioss/src/Ioss_ChangeSet.h
```diff
@@ -0,0 +1,101 @@
...
+
+#include <utility>
+#include <climits>
+#include <cstddef>
+#include <iostream>
+#include <sstream>
+#include <iomanip>
+#include <string>
+#include <functional>
+#include <map>
```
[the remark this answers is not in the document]
tokusanya: I usually rely on clang to help with cleanup of headers and stuff ... I'll look through these headers to see which are actually needed.

packages/seacas/libraries/ioss/src/Ioss_ChangeSet.h
```diff
@@ -0,0 +1,101 @@
...
+
+#include "Ioss_DatabaseIO.h"
+#include "Ioss_DBUsage.h"
+#include "Ioss_Region.h"
+#include "Ioss_ParallelUtils.h"
+
+#include <utility>
+#include <climits>
+#include <cstddef>
+#include <iostream>
```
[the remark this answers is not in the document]
tokusanya: Maybe not, I can check

packages/seacas/libraries/ioss/src/Ioss_ChangeSet.h
```diff
@@ -0,0 +1,101 @@
...
+
+#pragma once
+
+#include "Ioss_DatabaseIO.h"
+#include "Ioss_DBUsage.h"
+#include "Ioss_Region.h"
+#include "Ioss_ParallelUtils.h"
+
+#include <utility>
+#include <climits>
```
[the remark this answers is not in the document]
tokusanya: I'll check

packages/seacas/libraries/ioss/src/Ioss_ChangeSet.h
tokusanya: I'll add the header

packages/seacas/libraries/ioss/src/Ioss_ChangeSet.h
tokusanya: I wanted to use `enum class` but I use bitwise composition to figure out the supported formats so I left it as `enum`

packages/seacas/libraries/ioss/src/Ioss_ChangeSet.h
tokusanya: No problem

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.C
tokusanya: Yeah, the other version of `file_exists` makes use of it so I can convert this also

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.C
tokusanya: Sure, I can unify them

packages/seacas/libraries/ioss/src/Ioss_DynamicTopology.h
tokusanya: I can check

packages/seacas/libraries/ioss/src/exodus/Ioex_ChangeSet.h
tokusanya: Probably not...

packages/seacas/libraries/ioss/src/exodus/Ioex_ChangeSet.h
tokusanya: ChangeSets is an interface that is dependent on the implementation of the internal file format. Plain IOSS ChangeSet only handles the linear and cyclic versions. The Exodus ChangeSet handles the `group` implementation and other file formats such as catalyst which may implement their own versions.

packages/seacas/libraries/ioss/src/Ioss_Region.C
tokusanya: `fileControl` modifies those values as return values ... if you want it to be clearer as return values, I can look into this.

gsjaardema: But they are local to the function.  Do you want the regions values to be changed, or what is the intent?

packages/seacas/libraries/ioss/src/Ioss_ChangeSetFactory.h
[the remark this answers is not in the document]
tokusanya: Yeah, `clean` in this case does nothing ... I don't think even the Exodus version needs it. I can take it out for now

tokusanya:
@gdsjaar I have removed the draft status and marked it for review ... let me know if there is still anything required

packages/seacas/libraries/ioss/src/Ioss_DatabaseIO.h
```diff
@@ -246,63 +246,75 @@ namespace Ioss {
...
-     *  \param[in] group_name The name of the group to open.
+     *  \param[in] set_name The name of the set to open.
      *  \returns True if successful.
      */
-    bool open_group(const std::string &group_name)
+    bool open_internal_change_set(const std::string &set_name)
     {
       IOSS_FUNC_ENTER(m_);
-      return open_group_nl(group_name);
+      return open_internal_change_set_nl(set_name);
```
gsjaardema, line 258: Not sure every function with a lock (`IOSS_FUNC_ENTER(m_)`) needs a non-lock _nl version.  The _nl version is only needed if you need to call the function from another function which already holds a lock...

packages/seacas/libraries/ioss/src/exodus/Ioex_BaseDatabaseIO.C
```diff
@@ -507,9 +507,33 @@ namespace Ioex {
...
+    open_child_group_nl(0);
+  }
+
+  bool BaseDatabaseIO::supports_internal_change_set_nl()
+  {
+    return supports_group_nl();
   }
 
-  bool BaseDatabaseIO::open_root_group_nl()
+  bool BaseDatabaseIO::supports_group_nl() const
```
gsjaardema, line 520: Does this need a lock/no-lock version.  There should be no way tha another thread could change the format of this file while it is being queried.

(I probably need to get back in the thread-safe zen mode more for 100% confidence...)

packages/seacas/libraries/ioss/src/exodus/Ioex_BaseDatabaseIO.h
```diff
@@ -90,6 +90,94 @@ namespace Ioex {
     // database supports that type (e.g. return_value & Ioss::FACESET)
     IOSS_NODISCARD unsigned entity_field_support() const override;
 
+    IOSS_NODISCARD std::string get_internal_change_set_name() const override { return m_groupName; }
+
+    /** \brief Checks if a database type supports groups
+     *
+     *  \returns True if successful.
+     */
+    bool supports_group()
```
gsjaardema, line 99: Related to comment above -- is there a need to acquire a lock for this function

packages/seacas/libraries/ioss/src/exodus/Ioex_BaseDatabaseIO.h
```diff
@@ -90,6 +90,94 @@ namespace Ioex {
...
+     *
+     *  If the group_name begins with '/', it specifies the absolute path
+     *  name from the root with '/' separating groups.  Otherwise, the
+     *  group_name specifies a child group of the currently active group.
+     *  If group_name == "/" then the root group is opened.
+     *
+     *  \param[in] group_name The name of the group to open.
+     *  \returns True if successful.
+     */
+    bool open_group(const std::string &group_name)
```
gsjaardema, line 116: Similar.  Is there something that another thread coulf do to this file that requires us to get a lock...

And if so, is there another locked function that calls `open_group` which would require us to provide a no-lock version...

Probably something we need to go through all functions and verify 1) do they need to get a lock. and if so 2) does another locked function call this function (so we need both a lock and no-lock version).

packages/seacas/libraries/ioss/src/exodus/Ioex_ChangeSet.C
```diff
@@ -0,0 +1,160 @@
...
+                                                    unsigned fileCyclicCount) const
+  {
+    return new ChangeSet(db, dbName, dbType, fileCyclicCount);
+  }
+
+  // ========================================================================
+  ChangeSet::ChangeSet(Ioss::Region *region)
+      : Ioss::ChangeSet(region)
+  {
+    Ioss::ChangeSet::m_supportedFormats = (Ioss::CHANGE_SET_INTERNAL_FILES     |
```
gsjaardema, line 62: Should this be set just once as a member initializer in the .h file...

packages/seacas/libraries/ioss/src/exodus/Ioex_ParallelDatabaseIO.C
```diff
@@ -941,17 +949,20 @@ namespace Ioex {
     }
   }
 
-  void ParallelDatabaseIO::get_step_times_nl()
+  std::vector<double> ParallelDatabaseIO::internal_get_step_times_nl(bool setRegionTimeSteps)
```
gsjaardema, line 952: does this need to exist (_nl version) and does non-nl version need to acquiere a lock...

gsjaardema (commented):
Looks good.  I have some questions about the lock/no-lock functions.  

I'm fine with looking at this after merging or as part of this PR whichever you prefer.

packages/seacas/libraries/exodus/src/ex_inquire.c
gsjaardema: I think the only change here is to remove the `#if NC_HAS_HDF5`.  The `nc_inq_format` function call exists in non-hdf5-enabled builds and will return the format which can just be returned...

tokusanya: Yes, it is being being used as part of the check for whether an input Exodus database can actually support groups or not (as opposed to contains groups). Anything return value for the format other than netcdf4 is treated as false. I can remove the `#if` statement

gsjaardema: Yes, for this use case, but if it becomes part of the Exodus API, then it needs to work for all supported exodus formats.  Removing the #if should do that.  There are about 10 possible formats for the underlying netCDF file (most not commonly used yet).

packages/seacas/libraries/ioss/src/Ioss_ChangeSet.C
```diff
@@ -0,0 +1,382 @@
...
+
+  Ioss::FileInfo file = Ioss::FileInfo(filenameBase);
+  return file.parallel_exists(util.communicator(), message);
+}
+
+bool internal_decomp_specified(const Ioss::PropertyManager& props)
+{
+  bool internalDecompSpecified = false;
+
+  const std::string restartDecompMethod("RESTART_DECOMPOSITION_METHOD");
```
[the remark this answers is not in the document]
tokusanya: Ok, I'll add similar functionality to handle it the same way

packages/seacas/libraries/ioss/src/Ioss_ChangeSet.h
```diff
@@ -0,0 +1,104 @@
...
+
+  IOSS_NODISCARD unsigned supported_formats() const { return m_supportedFormats; }
+  IOSS_NODISCARD unsigned database_format() const { return m_databaseFormat; }
+
+  virtual void populate_change_sets(bool loadAllFiles = true);
+
+  IOSS_NODISCARD virtual DatabaseIO* open_change_set(unsigned index, Ioss::DatabaseUsage usage);
+  virtual void close_change_set(unsigned index);
+
+  IOSS_NODISCARD unsigned size() const { return m_changeSetNames.size(); }
```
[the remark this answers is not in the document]
tokusanya: Changed

packages/seacas/libraries/ioss/src/exodus/Ioex_BaseDatabaseIO.C
gsjaardema: Yes, that should be good.  If they make changes down at the exodus level, then exodus should take care of locking, so should only have to look at what happens at the IOSS level.

packages/seacas/libraries/ioss/src/exodus/Ioex_BaseDatabaseIO.h
tokusanya: Probably not but let me know what you'd like to do

packages/seacas/libraries/ioss/src/exodus/Ioex_BaseDatabaseIO.h
tokusanya: I believe `open_group` is called from a locked function but I'd like to go through them with you to confirm which needs to have lock/no-lock versions

packages/seacas/libraries/ioss/src/exodus/Ioex_ChangeSet.C
tokusanya: That is an inherited class member ... how would I set it in the .h file ? I can see setting it as part of the constructor initialization ...

packages/seacas/libraries/ioss/src/exodus/Ioex_ParallelDatabaseIO.C
tokusanya: Lines 1030 and 1035 are where this is called from and those are both _nl versions of calling functions. I would assume that since they are _nl functions, then there is no need to acquire a lock. Should I rename it to a non _nl version ?

gsjaardema: It is probably best to merge this and then go through and look at all the locks / _nl functions and see what is needed...  Too many locks shouldnt' affect anything (and thread-safe not run that much anyway)

packages/seacas/libraries/ioss/src/Ioss_ChangeSet.C
```diff
@@ -43,17 +43,39 @@ int file_exists( const Ioss::ParallelUtils &util,
   return file.parallel_exists(util.communicator(), message);
 }
 
-bool internal_decomp_specified(const Ioss::PropertyManager& props)
+std::string get_decomposition_property(const Ioss::PropertyManager &properties,
```
gsjaardema, line 46: That looks good.  At some point we maybe move to the Ioss::Utils class to avoid the code dupication, but good for now.

gsjaardema (approved):

gsjaardema:
@tokusanya Can you take a look at my latest commit and see if it makes sense.  (The ifdef'd code would be completely removed in the final; just left it there so easier to see what was there...

tokusanya:
@gdsjaar I think I see what is happening ... the code I put in to automatically open up the first child group in `finalize_file_open()` and `handle_output_file()` may have been interfering with those calls you #ifdef'd out

gsjaardema:
> @gdsjaar I think I see what is happening ... the code I put in to automatically open up the first child group in `finalize_file_open()` and `handle_output_file()` may have been interfering with those calls you #ifdef'd out

Yes, and when I put in code that zero'd out the file pointer if the change group call failed, it then caught the error which was always happening, but just using the old file pointer...

gsjaardema:
@tokusanya I am good with you merging whenever you feel that it is ready.  You should have permission to merge.

tokusanya:
> @tokusanya I am good with you merging whenever you feel that it is ready. You should have permission to merge.

Ok, I'm working on one last issue that might involve a quick commit

tokusanya:
@gsjaardema Can you help me resolve the conflict that popped up? I'm not sure which version is correct

gsjaardema:
> @gsjaardema Can you help me resolve the conflict that popped up? I'm not sure which version is correct

I will give it a try

tokusanya:
@gsjaardema I'm ready to merge but a little surprised that the tests did not automatically run after my last commit

gsjaardema:
Is this ready to merge?

tokusanya:
> Is this ready to merge?

Yes

gsjaardema:
> @gsjaardema I'm ready to merge but a little surprised that the tests did not automatically run after my last commit

It looks like they are currently running.

gsjaardema:
@tokusanya Looks like a couple build errors showing up.  Possibly a bad merge on Ioss_DynamicTopology.C and the `const` should be removed from Ioss_DatabaseIO.h, `reset_database()` ?

gsjaardema:
I'm not sure what happened on Ioss_DynamicTopology.C.  Maybe the "clang-format` workflow messed it up?

tokusanya:
> I'm not sure what happened on Ioss_DynamicTopology.C. Maybe the "clang-format` workflow messed it up?

Possibly ... I always have issues with that file every time I do a pull. The file formatting is completely messed up

tokusanya:
@gsjaardema Does it make sense to break up Ioss_DynamicTopology.C according to the classes in it? It's currently supporting 4 classes...

gsjaardema:
> @gsjaardema Does it make sense to break up Ioss_DynamicTopology.C according to the classes in it? It's currently supporting 4 classes...

Yes, that would probably be good.

gsjaardema:
@tokusanya Looks like builds succeeded.  Should this be merged now?

tokusanya:
> @tokusanya Looks like builds succeeded. Should this be merged now?

Just merged it :)

tokusanya:
@gsjaardema All that's needed now is a SEACAS snapshot into Sierra

gsjaardema:
@tokusanya Do you want to try the snapshot, or should I?

tokusanya:
@gsjaardema Is there a workflow for how to do the snapshot? I've never done it before ...

gsjaardema:
There is a list of the steps at https://github.com/sandialabs/seacas/wiki#developer-oriented-things under the bullet `Snapshot into Sierra`

I will do this snapshot since it has lots of other changes in addition to the Dynamic Toplogy changes, so more difficult to create a coherent set of commits...

PR #508, opened 2024-10-02
Add seacas patch to support Pamgen as a TriBITS external compliant pa…

tonnguy:
…ckage.

These changes would support Pamgen being used as a TriBITS external compliant package. This would enable installing Pamgem with Trilinos and using Pangem within seacas for example. 

This would resolve errors related to duplicate Pangem::all_libs targets being declared from trying to build trilinos and seacas with Pamgen.


externalTriBITSPamgem.patch
```diff
@@ -0,0 +1,78 @@
+diff --git a/TPLsList.cmake b/TPLsList.cmake
+index eab14e1676..8b38e12a75 100644
+--- a/TPLsList.cmake
++++ b/TPLsList.cmake
+@@ -68,7 +68,7 @@ SET( Seacas_TPLS_FINDMODS_CLASSIFICATIONS
+   DataWarp        "cmake/TPLs/"    SS
+   METIS           "cmake/TPLs/"    SS
+   ParMETIS        "cmake/TPLs/"    SS
+-  Pamgen          "cmake/TPLs/"    SS
++  Pamgen          "TRIBITS_PKG"    SS
```
gsjaardema, line 10: @bartlettroscoe Does this change allow a non-Trilinos build of SEACAS to find an externally installed (perhaps via Trilinos) version of PamGen?

tonnguy: Hi Greg, I don't know if it could support that in all cases since it relies on a TriBITS externally built Pamgen which is enabled by passing `Trilinos_ENABLE_Pamgen=ON`.

gsjaardema: Most likely the Pamgen would be TriBITs installed, so that should work...  Even if installed with spack, it is supplied by Trilinos.

Should be OK then.

gsjaardema (approved):

bartlettroscoe (commented):
Why put this in a patch file?  Why not just directly modify the files on 'master':

* [`TPLsList.cmake`](https://github.com/sandialabs/seacas/blob/master/TPLsList.cmake)
* [`cmake/TPLs/FindTPLPamgen.cmake`](https://github.com/sandialabs/seacas/blob/master/cmake/TPLs/FindTPLPamgen.cmake)

That will work for both of the use cases:

1. SEACAS and Pamgen are built as TriBITS packages from within the Trilinos TriBITS CMake project
2. Pamgen is built and installed as part of the Trilinos TriBITS CMake project and then SEACAS is built in its own CMake project that pulls in Pamgen as a TriBITS-compliant external package (which is what this issue is about?)

tonnguy:
I didn't realize that could work for both use cases, I can update this branch to apply the patch directly.

gsjaardema:
> Why put this in a patch file? Why not just directly modify the files on 'master':
> 
Good catch.  I only looked at the changes and didn't realize it was a patch file....

gsjaardema (approved):

bartlettroscoe (approved):
Looks great!  (Delete 60 lines and change one line and not lose any functionality. What is not to love about that :-))

tonnguy:
@gsjaardema I didn't have write access to this repository and can't merge. Would I need to be granted write access? Or if this can be merge without granting permissions that would work too.

gsjaardema:
Thanks for working on this.

tonnguy:
> Thanks for working on this.

Thank you, Greg and Ross! I'm very new to using TriBITS and seacas so this was a good learning experience also.

PR #543, opened 2024-11-13
Add numBlob to the parameters for the ex_init_params method.

mvlopri:
This was needed to initialize a new exodus file from inside of exodus.py using the init_params argument of the exodus class.

Add a test to ensure this capability does not regress and that ex_init_params works if new arguments need to be initalized in the future.

Address the tests not running due to the test and exodus.py script being placed in lib64. This still needs work to ensure it finds the libexodus.so library because SEACAS_LIBDIR is left blank during the CMake configure step.


gsjaardema (approved):
LGTM

PR #546, opened 2024-11-14
DTIO: Fix app-side memory leak

tokusanya:
Update to allow app the opportunity to clean up any data associated with a Region to prevent memory leaks

This happens when the Region is reset on a topology change


gsjaardema (approved):
LGTM

PR #565, opened 2024-12-19
DTIO update

tokusanya:
Make sure the correct value of stateCount is returned in the case of a dynamic topology change


gsjaardema (approved):
I think that looks good.  Not sure if any performance issues if gets called too much

PR #601, opened 2025-02-28
Use lib instead of lib64

mvlopri:
The change made to use CMAKE_INSTALL_LIBDIR is causing issues for the
installations due to the removal of the SEACAS_LIBDIR which was used
in the finding of libexodus.so for exodus.py. Using CMAKE_INSTALL_LIBDIR
will not currently work for this as it points to lib64 instead of lib
which is where libexodus.so is being installed.

This leaves us with a few choices with the revert of this being the easiest
option though likely not the option we want long term.


gsjaardema (approved):
Looks good to me for now.  As you mentioned, would be good to go back to gnu install dirs at some point.

PR #608, opened 2025-03-12
Fix usages of strcasecmp/strncasecmp

spyridon97:
Fix usages of strcasecmp/strncasecmp


spyridon97:
There seems to be no msvc ci machine. That's why these were not caught.

spyridon97:
@gdsjaar

gsjaardema:
We currently have the `SEACAS MSYS2 Build ` and then on https://ci.appveyor.com/project/gsjaardema/seacas we have a MINGW build.  

Will look into adding a MSVC ci build...

PR #627, opened 2025-04-02
Catalyst api 2 fix conn raw tests

tjotaha:
Fixed Catalyst tests for element blocks when writing
 connectivity_raw. Original code was scrambling the
 connectivity in this case and causing test failures
 with recent changes to io_shell --compare.


gsjaardema:
Thanks for the quick response.  Apologies for the broken io_shell --compare...

PR #632, opened 2025-04-08
patches CGNS to use find_package(ZLIB)

andrewcorrigan:
- Patches CGNS to call `find_package(ZLIB)`
- Fixes error shown building static libraries as reported in https://github.com/sandialabs/seacas/issues/631


gsjaardema:
Have you also submitted this patch to the CGNS github repository?  https://github.com/cgns/CGNS.

I'm fine merging this, but long-term would be best if it was included in the CGNS distribution.

andrewcorrigan:
let me give that a try.   I'm also second-guessing removing the `find_library` line.

andrewcorrigan:
I opened a PR here:   https://github.com/CGNS/CGNS/pull/834, with the old `find_library` call added back, just in case that's needed for compatibility on other systems.

PR #638, opened 2025-04-09
-DCMAKE_POSITION_INDEPENDENT_CODE:BOOL=ON

andrewcorrigan:
I need to build SEACAS  statically and link it into a shared library.  To support this, this PR proposes building with `-DCMAKE_POSITION_INDEPENDENT_CODE:BOOL=ON`.  It should have no effect when building SEACAS/Exodus/dependencies as shared libraries.

I saw this was already done for `TPL/faodel`


gsjaardema:
Thanks.  Can you also add to `cmake-config`

andrewcorrigan:
done

PR #662, opened 2025-05-22
Ejoin combine sets blocks

gsjaardema:
Start implementing the capability to join sets and blocks that have the same name on the input files into a single block/set on the output file.


gsjaardema (commented):
.

PR #669, opened 2025-06-04
EXODUS.PY: Fix num_blob() function

gsjaardema:
The summarize() function uses `num_blob()` which was returning an error.  I think `numBlob` was not being set up correctly in the exodus class init method.  This is one way to fix it.  Not sure if it is the best...


mvlopri:
I would say it might be better long term to perform this kind of data cleaning on set rather than get but this seems like a fine change to fix the issue. More importantly, was the summarize test not failing?

mvlopri:
I will see if I can restructure that test to prevent regression when I get a chance.

mvlopri (approved):

Issue #675, opened 2025-06-25
install error


jpf114:
CMake Error at packages/seacas/scripts/CMakeLists.txt:104 (IF):
  if given arguments:

    "include" "MATCHES" "^D:/Study/C++/vcpkg/packages/seacas_x64-windows/debug.*"

  Regular expression
  "^D:/Study/C++/vcpkg/packages/seacas_x64-windows/debug.*" cannot compile

-- Configuring incomplete, errors occurred!
RegularExpression::compile(): Nested *?+.

RegularExpression::compile(): Error in compile.

ninja: build stopped: subcommand failed.

PR #676, opened 2025-06-25
CI: Refactor so path containing regex character does not cause issues

gsjaardema:
If the install patch contains characters that are typically recognized by a regex library, usually something like "c++", then the code was failing while determining the include and lib path for the exodus python interface.

Refactor the code to fix the issue.  Also removed some EXCLUDE_FILES calls until TriBits is fixed...

Addresses #675


gsjaardema:
@mvlopri If you get a chance can you look at the changes to see if you think they are accomplishing the same thing as the original code.

PR #786, opened 2026-02-03
io_info: Fix ambiguous call to format_to.

vbrunini:
A clang 18 w/libstdc++ from gcc 14 build reported these calls as ambiguous between the fmt function and the std::format function. Explicitly adding the fmt:: qualification avoids the issue.


vbrunini:
@tokusanya

PR #790, opened 2026-03-02
Update `exomerge3` to work with (some) shell topologies

japlews:
Simply piggyback on `quad4` logic to support shells in `exomerge3`.


japlews:
@tokusanya I updated the MR with unit tests and handling for shell sides that were exposed by the tests.

PR #806, opened 2026-04-14
Polyhedral element support

anchen101:
adding Ioss files for defining polyhedral and polygonal elements. Updated several other files for polyhedral element support.


mrbuche:
@tokusanya these are the same changes from the other repo, just need your approval to run the workflows.

mrbuche:
@tokusanya I think it is requiring your approval to re-run the workflows every time we push, but we just added something that should fix the failures.

packages/seacas/libraries/ioss/src/elements/Ioss_Polygonal.C
```diff
@@ -0,0 +1,120 @@
...
+// ========================================================================
+namespace Ioss {
+  const char *Polygonal::name = "nsided";
+  class St_Polygonal : public ElementVariableType
+  {
+  public:
+    static void factory();
+
+  protected:
+    St_Polygonal() : ElementVariableType(Ioss::Polygonal::name, 13) {} // 13 is a magic number
```
tokusanya, line 24: If there is no specified number of nodes or quadrature points for this element type, why is 13 being used for the number of components?

packages/seacas/libraries/ioss/src/elements/Ioss_Polygonal.C
```diff
@@ -0,0 +1,120 @@
...
+}
+
+Ioss::Polygonal::Polygonal() : Ioss::ElementTopology(Ioss::Polygonal::name, Ioss::Polygonal::name)
+{
+  Ioss::ElementTopology::alias(Ioss::Polygonal::name, "Polygonal");
+  Ioss::ElementTopology::alias(Ioss::Polygonal::name, "nsided");
+}
+
+int Ioss::Polygonal::parametric_dimension() const { return 2; }
+int Ioss::Polygonal::spatial_dimension() const { return 2; }
```
tokusanya, line 54: While a surface is parametrically 2D, setting the spatial dimension here implies that this is only meant to be used in a 2D context. We have examples where there is distinction between the 2D and 3D versions of an element type e.g "edge"

packages/seacas/libraries/ioss/src/elements/Ioss_Polygonal.C
```diff
@@ -0,0 +1,120 @@
...
+
+Ioss::ElementTopology *Ioss::Polygonal::edge_type(int edge_number) const
+{
+  // edge_number == 0 returns topology for all edges if
+  // all edges are the same topology; otherwise, returns nullptr
+  // edge_number is 1-based.
+
+  assert(edge_number >= 0 && edge_number <= number_edges());
+  IOSS_ASSERT_USED(edge_number);
+  return Ioss::ElementTopology::factory("edge2");
```
tokusanya, line 119: As pointed out above, use of "edge2" as the edge_type implies that this is spatially 3D

packages/seacas/libraries/ioss/src/elements/Ioss_Polyhedral.C
```diff
@@ -0,0 +1,122 @@
...
+// ========================================================================
+namespace Ioss {
+  const char *Polyhedral::name = "nfaced";
+  class St_Polyhedral : public ElementVariableType
+  {
+  public:
+    static void factory();
+
+  protected:
+    St_Polyhedral() : ElementVariableType(Ioss::Polyhedral::name, 13) {} // 13 is a magic number
```
tokusanya, line 24: Again, what is the significance of this magic number? If it is to be used in both Polyhedral and Polygonal classes, we might as well make it a constexpr that is accessible for both

packages/seacas/libraries/ioss/src/exodus/Ioex_DatabaseIO.C
```diff
@@ -1354,6 +1366,13 @@ namespace Ioex {
 
       // See which connectivity options were defined for this block.
       // X -> Node is always defined.
+
+      // X -> Polyhedral or Polygonal?
+      if(type == "nfaced" || type.substr(0,6) == "nsided") {
```
tokusanya, line 1371: Maybe do the same substr() check for "nfaced"

packages/seacas/libraries/ioss/src/exodus/Ioex_DatabaseIO.C
```diff
@@ -1354,6 +1366,13 @@ namespace Ioex {
...
+        block->field_add(
+          Ioss::Field("entity_counts", block->field_int_type(), "Real[" + std::to_string(1) + "]", Ioss::Field::MESH));
```
tokusanya, line 1373: Is the implication here that there is only 1 entity per polyhedral/polygonal block?

packages/seacas/libraries/ioss/src/exodus/Ioex_DatabaseIO.C
```diff
@@ -2567,6 +2586,9 @@ namespace Ioex {
               }
             }
           }
+          else if (field.get_name() == "entity_counts") {
```
tokusanya, line 2589: This should probably be constrained to the scenario where the element block is polygonal/polyhedral else anyone could call this access for any element block which trickles down to the exodus polyhedral call

packages/seacas/libraries/ioss/src/exodus/Ioex_DatabaseIO.C
```diff
@@ -88,6 +88,18 @@ namespace {
     }
   }
 
+  void get_entity_count_data(int exoid, ex_entity_type type, ex_entity_id id, void* data)
```
tokusanya, line 91: Rename this so that it references polyhedra

packages/seacas/libraries/ioss/src/exodus/Ioex_DatabaseIO.C
```diff
@@ -2707,6 +2729,9 @@ namespace Ioex {
             // (eb_offset+1...eb_offset+1+my_face_count) to global face ids.
             get_map(EX_FACE_BLOCK).map_implicit_data(data, field, num_to_get, eb->get_offset());
           }
+          else if (field.get_name() == "entity_counts") {
```
tokusanya, line 2732: Same issue here as with element blocks

packages/seacas/libraries/ioss/src/elements/Ioss_Polygonal.C
mrbuche: Is this consistent with the other elements though? When I look at Quad4, Quad8, Tri3, Tri6, etc. they all have 2 for both.

packages/seacas/libraries/ioss/src/elements/Ioss_Polygonal.C
mrbuche: This one seems confusing too, we looked and Tri3 and Quad4 use `edge2` but Tri6 and Quad8 use `edge3`, so would the 2 or 3 here mean 2 or 3 nodes involved in the edge?

packages/seacas/libraries/ioss/src/elements/Ioss_Polygonal.C
mrbuche: We tried 0, but at some point something gets sized with this number, and 0 causes a terrible error. We weren't sure what number to use here, definitely open to suggestions.

packages/seacas/libraries/ioss/src/exodus/Ioex_DatabaseIO.C
anchen101: For this we want to add a field that holds 'the number of faces or nodes' that each element has. So we want 1 entity for each polyhedral/polygonal element not block. What would be the correct way to write this?

packages/seacas/libraries/ioss/src/Ioss_ElementTopology.C
```diff
@@ -111,6 +111,11 @@ Ioss::ElementTopology *Ioss::ElementTopology::factory(const std::string &type, b
     }
   }
 
+  std::string base2 = "nsided";
```
tokusanya, line 114: Where is the check for "nfaced" ?

tokusanya:
As a whole, it's a decent step in creating an Ioss wrapper around the polyhedral support in Exodus. Definitely needs unit tests showing

1. How to create polyhedral/polygonal element blocks
2. Characterizations on what to expect from these element types
3. How to read/write these element types in Ioss
4. Declaration and usage of Ioss::Field with these element blocks

mrbuche:
@tokusanya would you mind re-running the workflow, and maybe taking a peek at the changes when you have time?

mrbuche:
@tokusanya I think we fixed the failing tests just now, if you could re-run the workflows (or enable this PR to run them automatically somehow).

PR #809, opened 2026-04-21
Fix filesystem api

gsjaardema:
The use of `string_view` as the return type is ok for the linux implementation, but for the Mac, it is returning a value from the stack which is not good.  See #804 

Returning a `std::string` fixes this, but could potentially break a client using the function.  However, I don't think this function is being used much if at all and changing the API is hopefuly the easiest solution instead of deprecating and adding a new function...


gsjaardema:
Didn't realize other commits were added to this.  Will redo.

PR #810, opened 2026-04-21
EXODUS: Fix incorrect use of exi_redef

gsjaardema:
At the point that this function was calle in `ex_open_par`, the exodus `file_item` struct has not yet be initialized, so cannot use the `exi_redef` and have to use the `nc_redef`.

This should fix #805


gsjaardema:
@tokusanya This one is somewhat high priority.  Thanks

PR #811, opened 2026-04-21
Ioss fix filesystem type api

gsjaardema:
The `filesystem_type` function was returing a `string_view`. However, on the mac system, the value that was being returned was on the stack and left a dangling reference.  Changed the api to return a `std::string` instead.

This function probably isn't being used much or at all yet, so should not break anyone to change the public api; alternative is to depreacate this function and add a new good function, but it seems like overkill, but am also ok if that is the approach that is desired.

### DIFF
--- a/.github/workflows/build_aws.yml
+++ b/.github/workflows/build_aws.yml
@@ -30,7 +30,7 @@ jobs:
     steps:
 
       - name: Harden Runner
-        uses: step-security/harden-runner@f4a75cfd619ee5ce8d5b864b0d183aff3c69b55a # v2.13.1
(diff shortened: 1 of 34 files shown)


gsjaardema:
I guess this probably shouldn't be pushed since it contains more commits than just the fix to the filesystem api...
Most of the other changes are probably good except for the change to use my account on the docker push...

Forgot that my clone had some changes relative to sandialabs/seacas...

gsjaardema:
@tokusanya This does more than the title indicates.  It does fix #811, but also fixes a few other minor issues unrelated to that...  

I could probably separate into multiple PR if that makes it easier.

tokusanya:
@gsjaardema Do you still want to split this into separate PRs?

PR #813, opened 2026-04-27
Kluge to fix Ut_ioad test

gsjaardema:
Return value from type_string() is stored in retval to avoid errors.

Should not be needed, but it looks like something is causing an incorrect optimization or build...


gsjaardema:
This worked in local tests, but doesn't seem to affect the github actions tests...

PR #814, opened 2026-04-27
Update Intel oneAPI build setup in workflow

gsjaardema:
Removed installation of intel-oneapi-compiler-fortran and added setup for Intel oneAPI Fortran using fortran-lang/setup-fortran.

See if this fixes the currently failing intel-oneapi build...

### DIFF
--- a/.github/workflows/intel-build.yml
+++ b/.github/workflows/intel-build.yml
@@ -57,8 +57,9 @@ jobs:
       with:
         install-oneapi: true
         oneapi-version: '2025.3'
-        install-mpi: false
+        install-mpi: true
         install-mkl: false
+        install-classic: false
         
     - name: Install System dependencies
       shell: bash -l {0}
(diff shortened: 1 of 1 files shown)


gsjaardema:
@tokusanya This PR seems to get the intel build passing again. I used a provided action to install the intelOne compiler.  I hope it is a little more stable than the method we had been using before and maybe won't run into the space issue as often.

tokusanya:
> @tokusanya This PR seems to get the intel build passing again. I used a provided action to install the intelOne compiler. I hope it is a little more stable than the method we had been using before and maybe won't run into the space issue as often.

@gsjaardema Was the revert on the tmp reserve space increase in response to the previous fix that Mario put in where we increased it to 20Gb

gsjaardema:
No.  I think I tried increasing temp space even more and that gave an error so I reverted it back to what Mario had increased ot to.

PR #815, opened 2026-04-27
Initialize database in Ioad_BlockNames test case

gsjaardema:
I think that the database might need to be initialized to define variable types and database types...  Another shot in the dark...


gsjaardema:
@tokusanya This seems to fix the Ut_ioad issue and does some somewhat logical.

tokusanya:
> @tokusanya This seems to fix the Ut_ioad issue and does some somewhat logical.

I wonder why that would be the case here since the test is simply making use of pre-defined types and not actually creating databases like the other tests in the file

gsjaardema:
It think that the factory containing those types initialized     And I think also destructed when an initializer goes out of scope.  I need to look in more detail if that makes sense but haven't had a chance.

PR #816, opened 2026-04-28
Fix Use MODERN Cmake build

gsjaardema:
Add command to display CMakeCache.txt after configuration.

Want to see how the MODERN=YES is being resolved.

### DIFF
--- a/.github/workflows/build_variant.yml
+++ b/.github/workflows/build_variant.yml
@@ -57,7 +57,7 @@ jobs:
         uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
         with:
           path: ~/environments/${{ matrix.compiler }}-${{ matrix.hdf5 }}-${{ matrix.netcdf }}-${{ matrix.cgns }}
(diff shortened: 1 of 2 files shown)


gsjaardema:
@tokusanya Looks like this gets the MODERN build working again.  Not sure which changes are absolutely needed for this, but none of them are bad changes either.  

Probably want to update the NetCDF and CGNS versions on all workflows at some point in the future.  Only updated CGNS in this change.

The Intel failure and the ADIOS2 failure should be fixed by #814 and #815  so I think if these three are merged, all github tests should be passing....

tokusanya:
> @tokusanya Looks like this gets the MODERN build working again. Not sure which changes are absolutely needed for this, but none of them are bad changes either.
> 
> Probably want to update the NetCDF and CGNS versions on all workflows at some point in the future. Only updated CGNS in this change.
> 
> The Intel failure and the ADIOS2 failure should be fixed by #814 and #815 so I think if these three are merged, all github tests should be passing....

Thanks for looking at this @gsjaardema . Mario and I were banging our heads against a wall trying to resolve these issues

PR #822, opened 2026-06-12
Initial pass at adding attribute pass through to epu and unit tests

tokusanya:
Finish up by adding fix to epu output

  // We can't do nodal decomp testing yet
  //  [1] Nodal decomp of a mesh with elements doesn't work because Ioex_DatabaseIO fails on load: io_info exhibits this
  //  [2] Cannot use textmesh to generate a pure node mesh
  //  [3] Cannot test material properties on a mesh with no elements


packages/seacas/applications/ejoin/unit_tests/UnitTestEjoin.C
```diff
@@ -68,56 +68,18 @@ namespace {
...
-    }
   };
 
   TEST_F(EJoinTester, emptyTest) {}
 
-  TEST_F(EJoinTester, joinSingleHexMesh)
+  TEST_F(EJoinTester, joinSingleHexMeshWithMaterialProperties)
   {
     std::string inputFile  = "dummy.g";
     std::string outputFile = utest_util::unique_filename("singleHexJoin", "g");
```
tjotaha, line 85: Rather than creating a unique filename for each test, would it make sense to add something to the test fixture that creates a temporary directory before each test is run and then cleans it up afterwards?

packages/seacas/applications/utest_util/ElementPartition.h
tjotaha, line 1: Does the implementation of ElementPartition duplicate parts of the nem_slice application code? Maybe we should eventually refactor nem_slice to allow the functionality in ElementPartition to be accessed from there.

packages/seacas/applications/ejoin/unit_tests/UnitTestEjoin.C
tokusanya: That could work except that it doesn't quite fix the issue where two tests could create the same set of files in the directory that could clash with each other if the first test fails and ends up not cleaning up the temporary directory

packages/seacas/applications/utest_util/ElementPartition.h
tokusanya: Yes, there is some duplication since it is a subset of the nem_slice partitioning using Ioss instead of raw Exodus. I thought of how to organize this but not sure yet since we would probably need something similar to a nem_slice library ... otherwise the nem_spread unit test would have to link in files from the nem_slice directory which is weird

PR #825, opened 2026-06-23
IOSS: text mesh one dimensional meshes

tjotaha:
Allow beam_2 and beam_3 elements in one dimensional meshes.

### DIFF
--- a/packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
+++ b/packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
@@ -239,12 +239,12 @@ namespace Iotm {
 
         if (option.size() == 2) {
           m_parsedDimension = std::stoull(option[1]);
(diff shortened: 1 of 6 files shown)


packages/seacas/libraries/ioss/src/text_mesh/Iotm_TextMeshUtils.h
```diff
@@ -239,12 +239,12 @@ namespace Iotm {
 
         if (option.size() == 2) {
           m_parsedDimension = std::stoull(option[1]);
-          if (m_parsedDimension != 2 && m_parsedDimension != 3) {
-            std::ostringstream errmsg;
-            errmsg << "Error!  Parsed spatial dimension (" << m_parsedDimension
-                   << " not defined to be 2 or 3.";
-            m_errorHandler(errmsg);
-          }
+          // if (m_parsedDimension != 2 && m_parsedDimension != 3) {
```
tokusanya: Add extra conditional for 1D

tokusanya:
It looks like we may also need to disable the 1D textmesh unit test

PR #838, opened 2026-06-30
Extend TextMesh to enable usage of NODE topology

tokusanya:
This allows the creation of node only meshes

Added unit tests

Caveats include:
	Reference node can now be optional (see unit tests)
        Reference node must match the node entity identifier if provided
	Node entity can reference node connectivity from an element
	Reference block part is ignored for now ... may be used later

### DIFF
--- a/packages/seacas/libraries/ioss/src/unit_tests/UnitTestTextMesh.C
+++ b/packages/seacas/libraries/ioss/src/unit_tests/UnitTestTextMesh.C
@@ -777,17 +777,16 @@ namespace {
     if (get_parallel_size() != 1)
       return;
 
-    std::string         meshDesc        = "0,1,NODE,";
+    std::string meshDesc = "0,1,NODE,";
     EXPECT_NO_THROW(setup_text_mesh(meshDesc));
   }
-
 
   TEST_F(TestTextMesh, nodeWithNoReferenceVersion2)
   {
     if (get_parallel_size() != 1)
       return;
 
-    std::string         meshDesc        = "0,1,NODE";
+    std::string meshDesc = "0,1,NODE";
     EXPECT_NO_THROW(setup_text_mesh(meshDesc));
   }
 
@@ -796,7 +795,7 @@ namespace {
     if (get_parallel_size() != 1)
       return;
 
-    std::string         meshDesc        = "0,1,NODE,2";
+    std::string meshDesc = "0,1,NODE,2";
     EXPECT_THROW(setup_text_mesh(meshDesc), std::logic_error);
   }
 
@@ -805,8 +804,8 @@ namespace {
     if (get_parallel_size() != 1)
       return;
 
-    std::string         meshDesc        = "0,1,NODE,\n"
-                                          "0,5,NODE,";
+    std::string meshDesc = "0,1,NODE,\n"
+                           "0,5,NODE,";
     EXPECT_NO_THROW(setup_text_mesh(meshDesc));
   }
 
@@ -815,8 +814,8 @@ namespace {
     if (get_parallel_size() != 1)
       return;
 
-    std::string         meshDesc        = "0,1,NODE\n"
-                                          "0,5,NODE,";
+    std::string meshDesc = "0,1,NODE\n"
+                           "0,5,NODE,";
     EXPECT_NO_THROW(setup_text_mesh(meshDesc));
   }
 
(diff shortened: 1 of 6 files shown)


packages/seacas/libraries/ioss/src/unit_tests/UnitTestTextMesh.C
```diff
@@ -772,6 +772,72 @@ namespace {
...
+  }
+
+  TEST_F(TestTextMesh, threeNodes)
+  {
+    if (get_parallel_size() != 1)
+      return;
+
+    std::string meshDesc = "0,1,NODE,1\n"
+                           "0,2,NODE,2\n"
+                           "0,3,NODE,3";
```
alanw0: It seems like the id is specified twice on each line? Maybe nothing should be required after `NODE`.

packages/seacas/libraries/ioss/src/unit_tests/UnitTestTextMesh.C
```diff
@@ -772,6 +772,72 @@ namespace {
     EXPECT_THROW(setup_text_mesh(meshDesc), std::logic_error);
   }
 
+  TEST_F(TestTextMesh, nodeWithInvalidReference)
+  {
+    if (get_parallel_size() != 1)
+      return;
+
+    std::string meshDesc = "0,1,NODE,2";
```
alanw0: Can you still use `|dimension:2` if you want to specify spatial-dim, or is it hard-coded to 3?

tokusanya: Yes, you can even use `|dimension:1` if you like. We recently modified TextMesh to allow 1 dimensional meshes as part of the unit testing for nem_slice. That particular unit test fixture sets the dimension to 3 internally

packages/seacas/libraries/ioss/src/unit_tests/UnitTestTextMesh.C
tokusanya: In order to make use of the generalized parser that reads in the entity data info, you still need to specify the id of the reference node (which is the node itself in this case). There's all sort of checks such as making sure the number of specified connectivity nodes is the same as the number of nodes for the topology, etc. Adding in specialized parsing code for node topology that can ignore the reference node, etc will take a little more work for later. 

Also, if you specify a part for the node, it is currently being ignored. I'm undecided if to make a nodeset and just stick the node into this nodeset part or not. Also, the current implementation allows the node to reference a previously existing node connected to an element. I haven't quite decided if nodes defined this way should be part of a set of disconnected nodes only. Can you think of use cases in STK where you would want these nodes to be only disconnected nodes?

tjotaha: The first id before NODE is the node id? And the second id after NODE is the index into the coordinates?

tjotaha (approved):
Looks good.

packages/seacas/libraries/ioss/src/unit_tests/UnitTestTextMesh.C
alanw0: I would have guessed that those ids are arbitrary global ids, could have been `0,10000,NODE,10000`...

tokusanya: @tjotaha Yes, it's the global id. It is the same parser underneath that is being used for node and element entities. This means that in order to interpret the nodal connectivity associated with the node entity (in the same way as element entities), then the single global node connectivity id must be the same as the global id for the node entity. 

Not that this is different from using a PARTICLE topology which defines an element with a global id that is connected to the node in it's connectivity list. The sphere element id and the connected node id will typically be different in this case.

Also remember that once all the global node ids are collated, they are sorted and put into a list. This gives a one-to-one mapping between the provided coordinates data and the nodes. If there is a difference in the lengths of the two lists, then an error message is thrown

tokusanya: @alanw0 I just made another update to make the nodal connectivity for NODE topology optional. That way, the global id for the node is determined entirely by the listed id and not by the reference ids. If the user provides a reference id, then they must match

alanw0: Great, I like that.

PR #865, opened 2026-07-20
Update ElementPartition with beam element fix

tokusanya:
Also added some utility functions

### DIFF
--- a/packages/seacas/applications/utest_util/MeshFixture.h
+++ b/packages/seacas/applications/utest_util/MeshFixture.h
@@ -122,22 +122,6 @@ namespace utest_util {
                                                                         unsigned numProcs,
                                                                         bool     singleBlock);
 
-    /*template <size_t size> void clear_args(int &argc, const char *(&argv)[size])
-    {
-      argc = 0;
-
-      for (unsigned i = 0; i < size; i++) {
-        argv[i] = nullptr;
-      }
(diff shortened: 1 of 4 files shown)


jeffmauldin (approved):
First of all, I like the changes to the unit testing; that looks good.
The only question I have is: should ElementPartition() throw an exception when it can't find a node instead of returning -1? My expectation is that returning -1 better matches the rest of SEACAS but I thought I'd ask since it occurred to me.

packages/seacas/applications/utest_util/MeshFixture.h
```diff
@@ -122,22 +122,6 @@ namespace utest_util {
                                                                         unsigned numProcs,
                                                                         bool     singleBlock);
 
-    /*template <size_t size> void clear_args(int &argc, const char *(&argv)[size])
```
tjotaha, line 125: I forgot to remove this commented out code last commit.

tokusanya:
> First of all, I like the changes to the unit testing; that looks good. The only question I have is: should ElementPartition() throw an exception when it can't find a node instead of returning -1? My expectation is that returning -1 better matches the rest of SEACAS but I thought I'd ask since it occurred to me.

Could you let me know which function or line specifically this is?

tjotaha (approved):

PR #872, opened 2026-07-22
Expand testing for the node definition capability in TextMesh

tokusanya:
Confirm that node ids are independent of element ids


tokusanya:
@tjotaha @jeffmauldin

tjotaha (approved):
Looks good.

PR #883, opened 2026-07-31
Added 2D capability for ejoin

tokusanya:
Added unit test

### DIFF
--- a/packages/seacas/applications/ejoin/EJ_match.C
+++ b/packages/seacas/applications/ejoin/EJ_match.C
@@ -99,6 +99,12 @@ namespace {
 
   void fill_3D_coordinates(Ioss::NodeBlock *nb, std::vector<double> &coord)
   {
+    // Depending on the spatial dimension, the length of the coordinate vector
+    // varies. The usage of this returned vector is based on the assumption that
+    // it represents a 3D mesh. For lower dimensions, expand the vector and copy
+    // data values into the correct locations while setting other coordinate values
+    // to zero. This is done by doing a reverse loop so as not to overwrite values
+
     nb->get_field_data("mesh_model_coordinates", coord);
 
     auto db     = nb->get_database();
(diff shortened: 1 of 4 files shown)


tokusanya:
@tjotaha @jeffmauldin @davi0011

jeffmauldin (approved):
I'm looking at fill_3D_coordinates() and trying to convince myself that the 2 and 1 dimensional loops don't overwrite data from lower indexes before copying them into higher indexes.

Oh, now I see that we are looping with a decreasing  index, so we always use the lower index before we copy over it. Yes, this is fine.

tokusanya:
> I'm looking at fill_3D_coordinates() and trying to convince myself that the 2 and 1 dimensional loops don't overwrite data from lower indexes before copying them into higher indexes.
> 
> Oh, now I see that we are looping with a decreasing index, so we always use the lower index before we copy over it. Yes, this is fine.

Yes, I had to make sure the logic worked correctly by looping backwards

packages/seacas/applications/ejoin/EJ_match.C
```diff
@@ -97,6 +97,36 @@ namespace {
     }
   }
 
+  void fill_3D_coordinates(Ioss::NodeBlock *nb, std::vector<double> &coord)
```
tjotaha, line 100: So this function now handles the cases for 1D and 2D and returns the data in coord so it works like the 3D case when the function returns?

packages/seacas/applications/ejoin/unit_tests/UnitTestEjoin.C
```diff
@@ -114,4 +139,66 @@ namespace {
     unlink(outputFile.c_str());
   }
 
+  TEST_F(EJoinTester, join2DQuadMeshes)
```
tjotaha, line 142: Do we need a 1D test case?

tjotaha (approved):

packages/seacas/applications/ejoin/unit_tests/UnitTestEjoin.C
tokusanya: I though about this and wasn't even sure what a use case for this would be unless it was simply to merge two 1-D meshes like 1D beams. I guess it couldn't hurt seeing that we can now create 1D meshes with TextMesh

packages/seacas/applications/ejoin/EJ_match.C
tokusanya: Exactly

packages/seacas/applications/ejoin/unit_tests/UnitTestEjoin.C
tokusanya: Turns out the 1D test case exposed another place where the 3D coordinate conversion is needed

PR #890, opened 2026-08-04
Compsimhd 18606 ioss flush

tjotaha:
IOSS: exodus output disk flush
    
    Moved code to handle disk flushing for file per rank
    Exodus output into separate class with unit testing. Added
    new IOSS property FLUSH_ON_FIRST_OUTPUT that will trigger
    a flush to disk on the first time output is called.

### DIFF
--- a/packages/seacas/libraries/ioss/src/Ioss_FlushHandler.C
+++ b/packages/seacas/libraries/ioss/src/Ioss_FlushHandler.C
@@ -1,0 +1,70 @@
+// Copyright(C) 1999-2025 National Technology & Engineering Solutions
+// of Sandia, LLC (NTESS).  Under the terms of Contract DE-NA0003525 with
+// NTESS, the U.S. Government retains certain rights in this software.
+//
+// See packages/seacas/LICENSE for details
+
+#include "Ioss_FlushHandler.h"
+#include "Ioss_SerializeIO.h"
+
+namespace Ioss {
+  bool FlushHandler::doFlush(int state)
+  {
+    bool do_flush = false;
+    if (flushOnFirstOutput && isFirstOutput) {
+      do_flush = true;
+    }
+    else if (flushInterval == 1) {
+      do_flush = true;
+    }
+    else if (flushInterval < 0) {
+      do_flush = isGreaterThanFlushInterval(timeLastFlush, timeLastFlushInterval);
+    }
+    else if (flushInterval > 1) {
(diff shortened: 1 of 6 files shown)


packages/seacas/libraries/ioss/src/unit_tests/UnitTestFlushHandler.C
```diff
@@ -0,0 +1,135 @@
...
+
+  void TearDown() override {}
+};
+
+TEST_F(FlushTest, EmptyTest) {}
+
+TEST_F(FlushTest, DefaultFlushHandler)
+{
+  ASSERT_EQ(-1, fh.getFlushInterval());
+  ASSERT_TRUE(fh.getIsParallel());
```
tokusanya, line 58: Shouldn't this depend on compilation with MPI?

packages/seacas/libraries/ioss/src/Ioss_FlushHandler.C
```diff
@@ -0,0 +1,70 @@
...
+    }
+    return do_flush;
+  }
+
+  bool FlushHandler::allReduce(bool do_flush)
+  {
+    int iflush = do_flush ? 1 : 0;
+#ifdef SEACAS_HAVE_MPI
+    if (isParallel) {
+      if (Ioss::SerializeIO::isEnabled()) {
```
tokusanya, line 59: It would be great to have a unit test where we can turn off serialization and test this

tokusanya:
Very nice modularization and testing... I like it!

jeffmauldin:
As per our discussion, I built Sierra with this PR of SEACAS and verified that all of Tom's fixes work properly integrated. Specifically, if you set the property FLUSH_ON_FIRST_OUTPUT to 1 (or anything), the first output and all of its transients are flushed to disk and you get a dataset which has transient data (one timestep), while if you do not set the property that flush does not happen and if you kill the simulation after the first output you get the mesh with no transients as output. Additionally, the time delta now works properly. If you accept the default behavior which flushes if that time since last flush has been more than 10 seconds, Sierra output now behaves correctly. I tested our issue replicator so that FLUSH_ON_FIRST_OUTPUT was not set, but I made a second output happen more than 10 seconds after the first one and the flush gets triggered properly. When you interrupt the simulation after that second output you get a dataset that has 2 timesteps worth of transient data.
We can definitely accept this pull request as I have now additionally verified it with a (manual) Sierra integration test.

PR #918, opened 2026-09-04
Algebra block names

tokusanya:
Enable ALGEBRA to output element block, nodeset and sideset names


packages/seacas/scripts/run_algebra.in.py
```diff
@@ -0,0 +1,67 @@
...
+        print(f"Error: command file does not exist: {command_file}", file=sys.stderr)
+        sys.exit(1)
+
+    # Get the directory of the current script
+    script_dir = Path(__file__).resolve().parent
+
+    exe_dir = script_dir / '../applications/algebra'
+
+    # Convert to string and add it
+    #sys.path.insert(0, str(exe_dir))
```
tjotaha, line 40: Do we need this commented code?

packages/seacas/scripts/CMakeLists.txt
```diff
@@ -66,6 +66,21 @@ IF (${PROJECT_NAME}_ENABLE_SEACASConjoin)
   INSTALL(PROGRAMS ${CMAKE_CURRENT_BINARY_DIR}/pconjoin DESTINATION bin)
 ENDIF()
 
+ASSERT_DEFINED(${PROJECT_NAME}_ENABLE_SEACASAlgebra)
+IF (${PROJECT_NAME}_ENABLE_SEACASAlgebra)
+  find_package(Python COMPONENTS Interpreter)
+  IF (${Python_FOUND})
+    CONFIGURE_FILE(
```
tjotaha, line 73: Does the CONFIGURE_FILE command actually change anything in run_algebra.in.py, or is this used to get the file copied into the build directory and then installed?  An alternative approach is to use FILE and then INSTALL as shown below.

# 1. Copy from source to build directory during configuration
file(COPY my_script.py 
     DESTINATION ${CMAKE_CURRENT_BINARY_DIR}
     FILE_PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)

# 2. Install to the final system destination
install(PROGRAMS ${CMAKE_CURRENT_BINARY_DIR}/my_script.py 
        DESTINATION bin)

tjotaha (approved):

packages/seacas/scripts/run_algebra.in.py
tokusanya: Nope, I'll remove it

packages/seacas/scripts/CMakeLists.txt
tokusanya: I read that that it was to copy the file locally so that my faulty attempt at using STANDARD_INPUT to pipe in the command file would have worked. Since we are no longer going that route, I can supply the actual relative path for the command file instead of relying on a copy

tokusanya: I just re-read your comment and realized that I was responding to a different thing. I think using configure_file() is probably the right way to go for now in anticipation of customizing the script to wrap other executables

PR #926, opened 2026-09-14
Flush before entering define mode on a parallel file

### LINKED ISSUE #927: Variable names are corrupted when a parallel writer spans more than one node
This issue came up when investigating long-standing issue on my fracture code where "formatting" (adding variable names and truth table) a file on large number of nodes could get corrupted on some filesystems.

What is below is the Claude-generated explanation. I will open a PR with a trivial proposed fix right away

## Summary

`ex_put_variable_name()` data is silently corrupted if a later call grows the file
header, when the writing job uses more than one node. The run that writes the file
does not notice; every reader does.

The cause is not the name write. It is the data-section relocation that a later
header growth triggers: it reads previously written data back and rewrites it at a
higher offset, and on a multi-node job it can read bytes that another node wrote and
has not made visible. MPI-IO only guarantees that after a sync.

## Reproducer

`TestEXONameOverlap.c` (full source at the end of this issue) uses only exodus and MPI. It performs the normal
formatting sequence — nodal variable params, nodal names, zonal variable params,
zonal names, truth table — then closes the file and reads the names back.

```
$ srun -N1 -n8 ./TestEXONameOverlap $SCRATCH/nm.exo all
== 8 ranks on 1 node(s), mode=all
   nodal 3  wrote "Damage"  read back "Damage"
== PASS

$ srun -N2 -n8 ./TestEXONameOverlap $SCRATCH/nm.exo all
== 8 ranks on 2 node(s), mode=all
   nodal 3  wrote "Damage"  read back "Dama"   <-- WRONG
== FAIL

$ srun -N4 -n8 ./TestEXONameOverlap $SCRATCH/nm.exo all
   nodal 3  wrote "Damage"  read back "Da"     <-- WRONG
== FAIL
```

Same binary, same filesystem, same rank count. Only the number of nodes differs.
The tail of the name is left behind in the preceding 256-byte record, and how much
is lost scales with the node count.

## Why it is the relocation, not the write

Tracing MPI-IO with `MPI_File_set_view` tracked per file handle, every rank on every
node issues identical and correct requests, in the failing 2-node run and the passing
1-node run alike:

```
write_at_all abs=24688 bytes=15 data="Displacement_X"
write_at_all abs=24944 bytes=15 data="Displacement_Y"
write_at_all abs=25200 bytes=7  data="Damage"
```

Those are followed by

```
write_at_all abs=2560 bytes=23000    <- relocation after ex_put_variable_param(EX_ELEM_BLOCK)
write_at_all abs=2792 bytes=23768    <- relocation after ex_put_truth_table()
```

i.e. the names are written once and then relocated twice. Nothing above MPI-IO
computes a wrong offset.

Inside PnetCDF, `move_file_block()` moves the section in chunks, each rank handling
its own, and synchronises with `read / MPI_Barrier / write`. The barrier orders this
move's own reads against its own writes; it does not make writes issued *earlier*
visible to the rank that is about to read them.

## Workarounds that confirm it

```
mode      what it adds                                      N1     N2     N4
all       nothing (default)                                 PASS   FAIL   FAIL
barrier   MPI_Barrier after each ex_put_variable_name()      PASS   FAIL   FAIL
update    ex_update() before the header grows again          PASS   PASS   PASS
flushall  ex_update() before every header growth             PASS   PASS   PASS
root      rank 0 does all formatting via a serial handle      PASS   PASS   PASS
```

A barrier is not enough — it orders processes without making bytes visible. The flush
is what fixes it.

## Filesystem dependence — please read before trying to reproduce

**This does not reproduce on Lustre.** It was verified passing there on a separate
system. The execution path genuinely differs:

- ROMIO uses `ad_lustre` on Lustre and falls back to `ad_ufs` on BeeGFS, so
  `MPI_File_write_at_all` is a different implementation with different aggregator
  and file-domain logic.
- PnetCDF sizes its relocation chunk from the striping unit
  (`ncp->data_chunk = striping_unit > 0 ? striping_unit : PNC_DATA_MOVE_CHUNK_SIZE`),
  so the rank-to-byte mapping of the move differs; it also carries an explicit
  `PNCIO_FS_LUSTRE` vs `PNCIO_FS_UFS` distinction.
- Lustre's lock manager gives client cache coherence, so a write on one client
  becomes visible to another without an explicit sync. BeeGFS does not promise that.

So Lustre appears to hide the bug rather than not have it: the code relies on a
visibility guarantee MPI-IO does not provide, and Lustre happens to supply it.

Filesystem fingerprint where it reproduces (`MPI_File_get_info`):

```
romio_filesystem_type   UFS: Generic ROMIO driver for all UNIX-like file systems
striping_unit           0
striping_factor         0
cb_nodes                2          (one aggregator per node)
```

## Impact

The writing application does not see the problem, because the names it set are cached
in memory. Readers do. In our case (MEF90/vDef, via PETSc's ExodusII viewer) a restart
from a file written on 4 nodes fails with

```
[0]PETSC ERROR: Unexpected data in file
[0]PETSC ERROR: Could not find nodal or zonal variable Damage in exodus file.
```

while the same restart from a 1-node-written file succeeds. ParaView and other
post-processors see the mangled names too. Nothing warns at write time.

A fix is proposed in #926: flush in `exi_redef()`/`exi_persist_redef()`, the choke points through which every header-growing call enters define mode.

## Environment

- exodus from the SEACAS master bundled by PETSc, `PARALLEL_AWARE_EXODUS` enabled
- netCDF-c 4.10.1, PnetCDF master `8c05696d` (post-1.15.0); reproduces identically on
  PnetCDF 1.14.1, so it is independent of the 1.15.0 `move_file_block` overflow fix
- MVAPICH 4.0 (ROMIO), gcc 13.2.0, Linux 5.14.0 x86_64
- BeeGFS (BeeOND), CDF-2 (classic) exodus files

## Caveat on the diagnosis

The symptom and the cure are solid and reproducible. The mechanism is inferred. An
attempt to reproduce it with PnetCDF alone — repeated header growth with data written
in between, three write patterns, two data-section sizes — passes on 1, 2 and 4 nodes
in all combinations, so I could not isolate it below exodus. The missing-visibility
explanation fits every observation, but the flush also serialises, and it is possible
it masks something else. Treat the fix as verified and the cause as probable.

## Reproducer source

Build with a parallel-aware exodus (`ex_open_par`, i.e. `PARALLEL_AWARE_EXODUS`) plus MPI; no other dependencies:

```sh
mpicc -O0 -g -I$EXODUS/include TestEXONameOverlap.c -o TestEXONameOverlap \
      -L$EXODUS/lib -Wl,-rpath,$EXODUS/lib -lexodus
```

<details>
<summary>TestEXONameOverlap.c</summary>

```c
/* TestEXONameOverlap.c -- exodus variable names are corrupted when a parallel
 * writer spans more than one node.
 *
 * Declaring exodus result variables grows the file header, so PnetCDF relocates the
 * whole data section to a higher offset (ncmpio_enddef -> move_file_block). A normal
 * exodus formatting sequence does that several times in a row:
 *
 *   ex_put_variable_param(EX_NODAL, n)        header grows -> data section moved  (1)
 *   ex_put_variable_name (EX_NODAL, ...)      names written at their new offsets
 *   ex_put_variable_param(EX_ELEM_BLOCK, n)   header grows -> data section moved  (2)
 *   ex_put_variable_name (EX_ELEM_BLOCK, ...)
 *   ex_put_truth_table   (...)                header grows -> data section moved  (3)
 *
 * so the nodal names are written once and then *relocated twice* by move_file_block.
 * That is where they are damaged. Reading "Damage" back out of name_nod_var:
 *
 *   1 node    "Damage"     correct
 *   2 nodes   "Dama"       last bytes left behind in the previous 256-byte record
 *   4 nodes   "Da"
 *
 * The name writes themselves are correct. With MPI_File_set_view tracked per file
 * handle, every rank on every node issues exactly the same, correct requests --
 * identical in the passing 1-node run and the failing 2-node run:
 *
 *   write_at_all abs=24688 bytes=15 data="Displacement_X"
 *   write_at_all abs=24944 bytes=15 data="Displacement_Y"
 *   write_at_all abs=25200 bytes=7  data="Damage"
 *
 * followed later by  write_at_all abs=2560 bytes=23000  and  abs=2792 bytes=23768,
 * the two relocations of the block those names live in. Nothing above MPI-IO computes
 * a wrong offset; the damage happens while the section is being moved.
 *
 * SUSPECTED CAUSE (the cure is confirmed, the mechanism is not -- see the caveat
 * at the end). move_file_block() moves the section in chunks, each rank handling
 * its own chunk:
 *
 *     read  chunk at off_from      (MPI_File_read_at / pread)
 *     MPI_Barrier(ncp->comm)       "to prevent from one rank's write run faster
 *                                   than other's read"
 *     write chunk at off_to        (MPI_File_write_at / pwrite)
 *
 * The chunk partition depends only on rank and nprocs, which are the same in all
 * three runs above, so the arithmetic is not what differs. What differs is whether a
 * rank can *see* bytes written by a rank on another node. MPI-IO requires a
 * sync-barrier-sync sequence for that; a bare MPI_Barrier is not sufficient, and
 * across nodes the client caches are not coherent without it. On a single node the
 * page cache is shared, which is why one node always works.
 *
 * Adding sync-barrier-sync *before* the move loop fixes it. Rebuilt PnetCDF, same
 * binary and same nodes, only libpnetcdf swapped in with LD_PRELOAD:
 *
 *   stock                                   N2=FAIL  N4=FAIL
 *   sync-barrier-sync before the loop       N2=PASS  N4=PASS
 *   sync around the existing in-loop barrier N2=FAIL  N4=FAIL
 *   both                                    N2=PASS  N4=PASS
 *
 * so the entry sync is the whole fix: the hazard is seeing writes from *earlier*
 * phases, not ordering this move's own reads against its own writes, which is all
 * the existing barrier ever addressed. See pnetcdf-move_file_block-missing-sync.patch.
 *
 * With that patched PnetCDF, unmodified MEF90 vDef writes correct names on 2 and 4
 * nodes and restarts cleanly, so no change is needed above the library.
 *
 * CAVEAT on the mechanism: an attempt to reproduce this with PnetCDF alone failed --
 * see Tests/TestPnetCDFMoveStale.c, which exercises repeated header growth with data
 * written in between, in three write patterns and two data-section sizes, and passes
 * on 1, 2 and 4 nodes in all of them. So the missing-sync story explains the symptom
 * and the patch cures it, but the patch also flushes and serialises and may be
 * masking the real defect. Treat the fix as verified and the diagnosis as probable.
 *
 * Consistent with that, the result is aggregation-dependent -- three ROMIO settings
 * give three different wrong answers, and none of them is right:
 *
 *   default                    "Displacement_Yge"      "Dama"
 *   romio_cb_write disable     "Displacement_YDamage"  "Damage"   (stray extra copy)
 *   cb_nodes 1                 "Displacement_Y"        ""         (name lost entirely)
 *
 * Observed with PnetCDF 1.14.1 and with PnetCDF master (post-1.15.0) alike -- they
 * produce byte-identical corruption, so this is independent of the move_file_block()
 * 32-bit overflow that 1.15.0 fixed.
 *
 * Doing the formatting from a single rank on a serially opened file (mode "root"
 * below) avoids it completely, at no cost: this is metadata, there is nothing to
 * parallelise, and with one process there is nothing to synchronise.
 *
 * Why it matters downstream: the run that writes the file does not notice, because
 * the names it set are cached in memory. Any reader does. Restarting MEF90's vDef
 * from a file written on 4 nodes fails with
 *
 *   [0]PETSC ERROR: Unexpected data in file
 *   [0]PETSC ERROR: Could not find nodal or zonal variable Damage in exodus file.
 *
 * while the same restart from a 1-node-written file succeeds. ParaView and any other
 * post-processor see the mangled names too.
 *
 * Environment where this was observed:
 *   exodus + netCDF 4.10.1 + PnetCDF 1.14.1 / master, built by PETSc
 *   MVAPICH 4.0 (ROMIO), gcc 13.2.0, Linux 5.14.0 x86_64, BeeGFS (BeeOND)
 *
 * Build:
 *   mpicc -O0 -g -I$PETSC_DIR/$PETSC_ARCH/include TestEXONameOverlap.c \
 *         -o TestEXONameOverlap -L$PETSC_DIR/$PETSC_ARCH/lib \
 *         -Wl,-rpath,$PETSC_DIR/$PETSC_ARCH/lib -lexodus -lnetcdf -lpnetcdf -lhdf5_hl -lhdf5
 *
 * Run (needs a parallel filesystem, and ranks on more than one node to fail):
 *   srun -N1 -n8 ./TestEXONameOverlap $SCRATCH/nm.exo all     -> PASS
 *   srun -N2 -n8 ./TestEXONameOverlap $SCRATCH/nm.exo all     -> FAIL  "Damage" -> "Dama"
 *   srun -N4 -n8 ./TestEXONameOverlap $SCRATCH/nm.exo all     -> FAIL  "Damage" -> "Da"
 *   srun -N4 -n8 ./TestEXONameOverlap $SCRATCH/nm.exo root    -> PASS
 *
 * Arguments:
 *   <path>          file to create   (default ./TestEXONameOverlap.exo)
 *   all       every rank formats on the parallel handle (default, the failing case)
 *   barrier   as "all", plus MPI_Barrier after each ex_put_variable_name()
 *   update    as "all", plus ex_update() (nc_sync) after the names, before the
 *             next ex_put_variable_param() relocates them
 *   both      barrier and update
 *   flushall  as "all", plus ex_update() before EVERY call that grows the header
 *   root      rank 0 does all the formatting through a serial handle
 */

#include <mpi.h>
#include <exodusII.h>
#include <stdio.h>
#include <string.h>
#include <stdlib.h>
#include <unistd.h>

#define NVAR   3
#define NNODE  4000
#define NELEM  2000
static const char *names[NVAR]  = {"Displacement_X", "Displacement_Y", "Damage"};
static const char *znames[NVAR] = {"Stress_XX", "Stress_YY", "Stress_XY"};

int main(int argc, char **argv)
{
    int   rank, nprocs, namelen, i, exoid, err = 0;
    int   cpu_ws = sizeof(double), io_ws = sizeof(double);
    float version;
    char  host[MPI_MAX_PROCESSOR_NAME];
    const char *path = (argc > 1) ? argv[1] : "./TestEXONameOverlap.exo";
    const char *mode = (argc > 2) ? argv[2] : "all";
    int   root_only = !strcmp(mode, "root");
    int   do_barrier = !strcmp(mode, "barrier") || !strcmp(mode, "both");
    int   do_update  = !strcmp(mode, "update")  || !strcmp(mode, "both");
    int   do_flushall = !strcmp(mode, "flushall");   /* flush before EVERY header growth */

    MPI_Init(&argc, &argv);
    MPI_Comm_rank(MPI_COMM_WORLD, &rank);
    MPI_Comm_size(MPI_COMM_WORLD, &nprocs);
    MPI_Get_processor_name(host, &namelen);

    /* how many distinct nodes are we on? that is what decides the outcome */
    {
        char *all = malloc((size_t)nprocs * MPI_MAX_PROCESSOR_NAME);
        MPI_Gather(host, MPI_MAX_PROCESSOR_NAME, MPI_CHAR,
                   all, MPI_MAX_PROCESSOR_NAME, MPI_CHAR, 0, MPI_COMM_WORLD);
        if (rank == 0) {
            int n = 0, j, k;
            for (j = 0; j < nprocs; j++) {
                for (k = 0; k < j; k++)
                    if (!strcmp(all + (size_t)j * MPI_MAX_PROCESSOR_NAME,
                                all + (size_t)k * MPI_MAX_PROCESSOR_NAME)) break;
                if (k == j) n++;
            }
            printf("== %d ranks on %d node(s), mode=%s, file %s\n",
                   nprocs, n, mode, path);
        }
        free(all);
    }

    /* rank 0 creates the file and writes a mesh: exodus only allows writing geometry
     * at creation, and the mesh matters -- it gives the file a data section, so that
     * declaring the variables below grows the header and makes pnetcdf shift the data
     * section, which is what puts the name records where they end up being split. */
    if (rank == 0) {
        int     nn = NNODE, ne = NELEM, j;
        double *x = malloc(nn * sizeof(double)), *y = malloc(nn * sizeof(double));
        int    *conn = malloc(3 * ne * sizeof(int));
        unlink(path);
        exoid = ex_create(path, EX_CLOBBER, &cpu_ws, &io_ws);
        if (exoid < 0) { fprintf(stderr, "ex_create failed\n"); MPI_Abort(MPI_COMM_WORLD, 1); }
        ex_put_init(exoid, "TestEXONameOverlap", 2, nn, ne, 1, 0, 0);
        for (j = 0; j < nn; j++) { x[j] = (double)j; y[j] = 0.5 * j; }
        ex_put_coord(exoid, x, y, NULL);
        ex_put_block(exoid, EX_ELEM_BLOCK, 1, "TRI3", ne, 3, 0, 0, 0);
        for (j = 0; j < ne; j++) {
            conn[3 * j + 0] = (j % nn) + 1;
            conn[3 * j + 1] = ((j + 1) % nn) + 1;
            conn[3 * j + 2] = ((j + 2) % nn) + 1;
        }
        ex_put_conn(exoid, EX_ELEM_BLOCK, 1, conn, NULL, NULL);
        ex_close(exoid);
        free(x); free(y); free(conn);
    }
    MPI_Barrier(MPI_COMM_WORLD);

    /* everybody reopens in parallel and declares the variables */
    exoid = ex_open_par(path, EX_WRITE, &cpu_ws, &io_ws, &version,
                        MPI_COMM_WORLD, MPI_INFO_NULL);
    if (exoid < 0) { fprintf(stderr, "[%d] ex_open_par failed\n", rank); MPI_Abort(MPI_COMM_WORLD, 1); }
    ex_put_variable_param(exoid, EX_NODAL, NVAR);

    if (!root_only) {
        /* exactly what MEF90EXOFormat does, on the parallel handle from every rank */
        for (i = 0; i < NVAR; i++) {
            ex_put_variable_name(exoid, EX_NODAL, i + 1, names[i]);
            /* candidate workaround A: barrier after each name write */
            if (do_barrier) MPI_Barrier(MPI_COMM_WORLD);
        }
        /* candidate workaround B: flush the names to the file before the next
         * ex_put_variable_param() grows the header and relocates them.
         * ex_update() is nc_sync(), i.e. MPI_File_sync on both handles. */
        if (do_update) { ex_update(exoid); MPI_Barrier(MPI_COMM_WORLD); }
        /* declaring the zonal variables grows the header again, so pnetcdf relocates
         * the data section -- including the nodal names just written above */
        if (do_flushall) { ex_update(exoid); MPI_Barrier(MPI_COMM_WORLD); }
        ex_put_variable_param(exoid, EX_ELEM_BLOCK, NVAR);
        for (i = 0; i < NVAR; i++) ex_put_variable_name(exoid, EX_ELEM_BLOCK, i + 1, znames[i]);
        /* and the truth table defines one variable per block per zonal variable,
         * growing the header a third time and relocating everything once more */
        {
            int *tab = malloc(NVAR * sizeof(int));
            for (i = 0; i < NVAR; i++) tab[i] = 1;
            if (do_flushall) { ex_update(exoid); MPI_Barrier(MPI_COMM_WORLD); }
            ex_put_truth_table(exoid, EX_ELEM_BLOCK, 1, NVAR, tab);
            free(tab);
        }
        ex_close(exoid);
    } else {
        /* THE FIX: close the parallel handle and let one rank do all of the
         * formatting through a serial handle, then reopen */
        ex_close(exoid);
        MPI_Barrier(MPI_COMM_WORLD);
        if (rank == 0) {
            int *tab = malloc(NVAR * sizeof(int));
            exoid = ex_open(path, EX_WRITE, &cpu_ws, &io_ws, &version);
            for (i = 0; i < NVAR; i++) ex_put_variable_name(exoid, EX_NODAL, i + 1, names[i]);
            if (do_flushall) { ex_update(exoid); MPI_Barrier(MPI_COMM_WORLD); }
        ex_put_variable_param(exoid, EX_ELEM_BLOCK, NVAR);
            for (i = 0; i < NVAR; i++) ex_put_variable_name(exoid, EX_ELEM_BLOCK, i + 1, znames[i]);
            for (i = 0; i < NVAR; i++) tab[i] = 1;
            ex_put_truth_table(exoid, EX_ELEM_BLOCK, 1, NVAR, tab);
            ex_close(exoid);
            free(tab);
        }
    }
    MPI_Barrier(MPI_COMM_WORLD);

    /* rank 0 reads the names back from the closed file */
    if (rank == 0) {
        char got[NVAR][MAX_STR_LENGTH + 1];
        exoid = ex_open(path, EX_READ, &cpu_ws, &io_ws, &version);
        for (i = 0; i < NVAR; i++) {
            memset(got[i], 0, sizeof(got[i]));
            ex_get_variable_name(exoid, EX_NODAL, i + 1, got[i]);
            if (strcmp(got[i], names[i])) err++;
            printf("   nodal %d  wrote \"%s\"  read back \"%s\"%s\n",
                   i + 1, names[i], got[i], strcmp(got[i], names[i]) ? "   <-- WRONG" : "");
        }
        for (i = 0; i < NVAR; i++) {
            memset(got[i], 0, sizeof(got[i]));
            ex_get_variable_name(exoid, EX_ELEM_BLOCK, i + 1, got[i]);
            if (strcmp(got[i], znames[i])) err++;
            printf("   zonal %d  wrote \"%s\"  read back \"%s\"%s\n",
                   i + 1, znames[i], got[i], strcmp(got[i], znames[i]) ? "   <-- WRONG" : "");
        }
        ex_close(exoid);
        printf("== %s\n", err ? "FAIL" : "PASS");
    }
    MPI_Bcast(&err, 1, MPI_INT, 0, MPI_COMM_WORLD);
    MPI_Finalize();
    return err ? 1 : 0;
}

```

</details>

bourdin:
Fixes #927

Fixes the corruption of variable names — and any other metadata written between two
header growths — when a parallel exodus writer spans more than one node.

I have a self-contained reproducer (exodus + MPI only, no other dependencies) that
fails deterministically at 2 and 4 nodes and passes at 1; happy to attach it here or
open it as a separate issue, whichever you prefer.

## The problem

Adding or resizing metadata grows the file header, and leaving define mode makes the
underlying parallel netCDF layer relocate the whole data section to a higher offset.
That relocation reads data back and rewrites it elsewhere, so it has to see
everything written so far — including writes issued by other MPI ranks.

MPI-IO does not guarantee that. A process sees another process's writes only after a
sync; a bare `MPI_Barrier` is not sufficient. So a normal formatting sequence

```c
ex_put_variable_param(exoid, EX_NODAL, n);        /* header grows -> section moved */
ex_put_variable_name (exoid, EX_NODAL, i, name);  /* names written                 */
ex_put_variable_param(exoid, EX_ELEM_BLOCK, m);   /* header grows -> section moved */
ex_put_truth_table   (exoid, ...);                /* header grows -> section moved */
```

loses part of the names, because the second and third relocations read them back
from a rank that may not see what another node wrote.

## The change

`exi_redef()` and `exi_persist_redef()` are the single choke points through which
every header-growing call enters define mode. Flush there when the file is parallel.

```c
static void exi_sync_before_redef(int exoid, const struct exi_file_item *file)
{
  if (file->is_parallel) {
    nc_sync(exoid);
  }
}
```

`nc_sync()` is collective, and every rank reaches `exi_redef()` together, so calling
it here is safe. It is a no-op for serial files, and on a parallel file it only costs
one flush per header growth — an operation that is already rewriting the entire data
section, so the relative cost is negligible.

## Verification

Tested by interposing `exi_redef`/`exi_persist_redef` with exactly these semantics
(both are exported and reached through the PLT, so the interposition exercises the
real call sites), against that reproducer:

```
                  N1     N2     N4
stock exodus      PASS   FAIL   FAIL
patched           PASS   PASS   PASS
```

Independently, the equivalent flush applied from the calling application —
`ex_update()` after the names and before the next header growth — also fixes it, on
1, 2 and 4 nodes, which is the workaround we have deployed meanwhile.

## Notes for reviewers

- **This will not reproduce on Lustre.** Verified passing there. Lustre's lock
  manager gives client cache coherence, ROMIO uses `ad_lustre` rather than the
  generic `ad_ufs` fallback, and PnetCDF sizes its relocation chunk from the striping
  unit, so both the visibility guarantee and the rank-to-byte mapping of the move
  differ. Reproducing needs a filesystem without a ROMIO driver and without coherent
  client caching; ours is BeeGFS/BeeOND.

- An alternative fix is possible one layer down, in PnetCDF's `move_file_block()`,
  which synchronises its read and write phases with a bare `MPI_Barrier`. Adding
  sync-barrier-sync before its move loop also cures the symptom (tested). I am
  proposing it here instead because the flush is correct on every filesystem and
  costs nothing where it was not needed, whereas a PnetCDF-internal sync would be
  paid by everyone including the systems that never had the problem. If you consider
  this PnetCDF's bug rather than exodus's, that patch is available.

- The mechanism is inferred, not proven: an attempt to reproduce with PnetCDF alone
  passed in all 18 combinations tried, so I could not isolate it below exodus. The
  fix is verified; the explanation is the best one consistent with the evidence.


bourdin:
Analysis, reproducer and filesystem caveats are in #927.

PR #928, opened 2026-09-14
Add EX_OPT_NONBLOCKING: non-blocking parallel variable output via Pne…

bourdin:
This is another Claude-generated PR adding non-blocking IO when using pnetcdf. The whole idea was suggested by @wkliao, from pnetcdf.

My first experiments show a speed-up of 2x to 6x on a lustre filesystem (Digital Research Alliance of Canada rorqual system).

Disclaimer: I don't pretend to understand the code, but what I do understand makes sense.

----------- CLAUDE-gernerated PR begins --------------

Exodus writes variable values with nc_put_vara_{float,double}(), which netcdf-c turns into the blocking collective ncmpi_put_vara_*_all() for a PnetCDF-backed file. That is one collective MPI-IO operation per (variable x entity block x time step), so the collective count grows with the number of blocks and fields rather than with the data volume.

This adds an opt-in mode that posts those writes with ncmpi_iput_vara_* and completes them with a single ncmpi_wait_all() per flush:

    ex_set_option(exoid, EX_OPT_NONBLOCKING, 1);   /* collective */
    ...
    ex_put_partial_var(...);                       /* unchanged */
    ex_flush(exoid);                               /* new; per time step */

Converted entry points: ex_put_partial_var() and exi_put_partial_nodal_var(). Everything else is untouched.

Applicability. The feature is gated three ways and is safe to enable on any file:

  - compile time, #if NC_HAS_PNETCDF;
  - per file, nc_inq_format_extended() == NC_FORMATX_PNETCDF, so netCDF-4/HDF5 parallel files and serial files keep the blocking path;
  - a layout self-check on the recovered netcdf-c ncid (below).

ex_set_option(EX_OPT_NONBLOCKING, 1) succeeds on every kind of file and simply does not engage where it does not apply, so callers need not know which backend they were given.

Reaching PnetCDF. netcdf-c exposes no non-blocking API -- libsrcp/ncpdispatch.c only ever calls the blocking ncmpi_put_vara_*[_all], and there is no dispatch slot for a non-blocking variant. The PnetCDF file id is therefore recovered through netcdf-c's exported NC_check_id() using a layout-compatible prefix of struct NC (declared in netcdf-c's non-installed include/nc.h, stable for many releases). The resolver verifies the recovered ext_ncid against the id it passed in, so if netcdf-c ever reorders those members the check fails and exodus falls back to blocking: a layout change costs performance, never correctness. A small nc_inq_pnetcdf_ncid() accessor upstream in netcdf-c would make this unnecessary.

Buffer ownership. ncmpi_iput_vara_*() does not copy the caller's buffer, but the exodus API promises that var_vals may be freed or reused as soon as the put returns, and callers rely on it -- PETSc's DMPlex writer hands over a temporary subvector's array and destroys it immediately. Values are therefore copied into a per-file staging arena and the request points at the copy. The arena is a chunk list that is only ever appended to; it is never realloc'd while requests are outstanding, since that would move bytes PnetCDF holds pointers to. ncmpi_bput_* was considered and rejected: it fails with NC_EINSUFFBUF when its attached buffer fills, and exodus cannot size that buffer up front, whereas it can always flush its own arena and continue.

Flushing is required for correctness, not just hygiene. PnetCDF's request record holds a raw NC_var pointer and a file offset, redef/enddef may move both, and neither flushes; close *cancels* rather than completes pending put requests, with only a printf warning. Pending requests are therefore flushed in exi_redef(), exi_persist_redef(), ex_update() and ex_close(), plus the new public ex_flush(). ncmpi_wait_all() is collective, so every process must reach each flush; to guarantee that even for a process that posted nothing, the PnetCDF id is resolved when the option is enabled rather than lazily on first write.

One netcdf-c interaction worth recording: netcdf-c leaves PnetCDF files in independent data mode (ncpdispatch.c puts them there at open and after every enddef), where the collective ncmpi_wait_all() is illegal and returns NC_EINDEP. The flush therefore calls nc_var_par_access(exoid, NC_GLOBAL, NC_COLLECTIVE) first, so netcdf-c performs the mode switch itself and its record of the access mode stays consistent with PnetCDF's. Calling ncmpi_end_indep_data() directly would desynchronize the two libraries.

No build system change is required. The two new files are named ex_*.c to match the existing FILE(GLOB SOURCES src/ex_*.c), and the Netcdf TPL already carries pnetcdf where PnetCDF support is configured.

Testing. Output is byte-identical between pristine exodus, patched with the option off, and patched with the option on. Seventeen of the bundled tests (testwt, testrd, testwt1, testrd1, testwtd, testrdd, testrd-nfaced, CreateEdgeFace, ReadEdgeFace, testwt-zeroe, testwt-zeron, testwt-one-attrib, test-empty, testwt-partial, testwt_clb, testwt-assembly, testrd-assembly) produce identical output before and after. An LD_PRELOAD interposer counting PnetCDF entry points confirms the path is genuinely taken: ncmpi_iput_vara_ double 0 -> 12 and ncmpi_put_vara_double_all 15 -> 3 on a 3-step, 4-variable, 2-block case. Enabling the option on netCDF-4/HDF5-parallel, serial classic and serial netCDF-4 files leaves it disabled, as intended.

Performance, on BeeGFS, 200k nodes, 20 steps, ~500 MiB, varying only how finely the payload is split into collective writes:

  blocks x elem/blk, evars   put calls   blocking   non-blocking   speedup
  4 x 160000, 4                    400    5.265 s        4.947 s     1.06x
  32 x 20000, 4                   2640    7.590 s        5.137 s     1.48x
  128 x 5000, 4                  10320   14.151 s        5.992 s     2.36x
  128 x 5000, 12                 30800   40.172 s       15.935 s     2.52x

Across nodes (16 ranks/node, 128 blocks) the gap roughly doubles, because each collective now crosses the interconnect: 2.49x on 1 node, 3.35x on 2, 4.20x on 4, 3.90x on 8. Blocking time drifts up with node count while non-blocking falls, so the curves diverge. On a single node with few large writes the staging copy is not repaid and non-blocking can be ~20% slower on a RAM disk, which is why the option defaults to off.

Flush every time step. Buffering across steps monotonically destroys the benefit (1.49x at flush-per-step down to 0.98x at flush-only-at-close) and grows the staging arena, so ex_flush() per step is both the cheapest in memory and the fastest.


wkliao:
Hi, @bourdin 
I am glad to hear that using PnetCDF nonblocking APIs results in a noticeable performance improvement.
Could you let me know what PnetCDF version was used in this eveluation?

bourdin:
Hi,

I used 1.15.1 + zero size patch

tokusanya:
This looks promising and may address some recent performance issues we were looking into

bourdin:
Great. What is the release schedule for seacas? I am trying to figure out if there is any way to get this in before the next petsc release

tokusanya:
Currently working on getting the code base into a consistent state for a TAG

jeffmauldin:
This does indeed look promising. I was reading through the description and it's complicated but I get what we are driving at. A few questions: (1) if I understand correctly, there is the creation (and release) of additional temporary buffers and associated copying to maintain the exodus interface and also make it work correctly. The performance of the extra copying doesn't scare me, as I know in-memory copies to be blazingly fast. However, is this going to significantly boost the memory usage of the data writing, especially when you have very large data and a large number of MPI ranks, (2) can this be (or is it already) implemented in such a way that we can toggle whether we are using this new or old behavior via a flag in CMakeLists.txt and/or #defines in the code?

bourdin:
1. That is correct. It indeed doubles the data usage since we copy that data that is to be written in a temporary buffer, but I don't see a way around this. Otherwise, one may free the data we are trying to write before it is actually written to disk. That said, it only grows with the number of MPI rank *per node*, so for a code that scales (strongly) well, that is not going to be a problem. If not, there is always 2.
2. That is already in: the default behaviour is still to use blocking functions, but if the file used pnetcdf and one calls ex_set_option(exoid, EX_OPT_NONBLOCKING, 1), then we use non-blocking IO. I think that this is more flexible than a compile-time flag

bourdin:
Thinking about it. If the buffer allocation is an issue, I can rewrite the function without it, noting that buffers should not be deallocated without calling ex_flush first. 

Let me know what you think

bourdin:
I've thought about a zero-copy mode, but I am worried that it is going to be much more fragile. Is the current implementation with a buffer copy preventing the PR to be merged? It would be nice if it could go I before the next PETSc point release.

PR #933, opened 2026-09-16
Add inherited surface_touching_blocks() method to DashSurfaceMesh

tokusanya:
Resolves sierra spack integration unit test failures


packages/seacas/libraries/ioss/src/generated/Iogn_DashSurfaceMesh.C
```diff
@@ -138,6 +138,23 @@ namespace Iogn {
     }
   }
 
+  Ioss::NameList DashSurfaceMesh::sideset_touching_blocks(int64_t setId) const
+  {
+    Ioss::NameList touchingBlocks;
+
+    switch (setId) {
+    case 1:
+      touchingBlocks.push_back("block_2");
```
tjotaha, line 147: For the dash surface mesh, side set 1 is always touching block_2 and side set 2 is always touching block_1? Kind of like two halves of a book?

